PR #3, opened 2020-03-10
CI Build

jframe:
Add the CI build and gradlew so we have a basic build for the project.


gradle/versions.gradle
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2018 ConsenSys AG.
```
rain-on: 2020?

jframe: done

PR #5, opened 2020-03-11
Bls keystore

jframe:
Copy of the bls keystore implementation from Teku


bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/Cipher.java
```diff
@@ -0,0 +1,76 @@
...
+  public Cipher(
+      @JsonProperty(value = "function", required = true) final CipherFunction cipherFunction,
+      @JsonProperty(value = "params", required = true) final CipherParam cipherParam,
+      @JsonProperty(value = "message", required = true) final Bytes message) {
+    this.cipherFunction = cipherFunction;
+    this.cipherParam = cipherParam;
+    this.message = message;
+  }
+
+  public Cipher(final CipherFunction cipherFunction, final Bytes ivParam) {
```
rain-on: new function?

bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/SCryptParam.java
```diff
@@ -0,0 +1,136 @@
...
+      @JsonProperty(value = "p", required = true) final int p,
+      @JsonProperty(value = "r", required = true) final int r,
+      @JsonProperty(value = "salt", required = true) final Bytes salt) {
+    super(dklen, salt);
+    this.n = n;
+    this.p = p;
+    this.r = r;
+  }
+
+  /**
```
rain-on: new?

gradle/versions.gradle
```diff
@@ -13,5 +13,38 @@
 
 dependencyManagement {
   dependencies {
+    dependency 'com.fasterxml.jackson.core:jackson-databind:2.10.1'
+
+    dependency 'com.google.errorprone:error_prone_annotation:2.3.4'
```
rain-on: did you want all of this coming in (because we probably don't need all of it - eg auto-service

bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/Cipher.java
jframe: Yeah I copied this validator command branch. That has been merged now so this is the same as master

bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/SCryptParam.java
jframe: Yeah I copied this validator command branch. That has been merged now so this is the same as master

gradle/versions.gradle
jframe: cleaned up, should only have what is needed now

PR #6, opened 2020-03-12
Rework Hashicorp DSL

rain-on:
Allows any key to be stored in the hashicorp vault.
Allows users of the HashicorpNode to get access to the underlying SelfSignedCertificate, such that it can be formatted appropriately, then used to construct a HashicorpConnection.


rain-on:
splitting into smaller PRs

PR #8, opened 2020-03-12
SelfSigned Certificate is now in memory

rain-on:


acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpNode.java
```diff
@@ -32,29 +37,37 @@ private HashicorpNode(final DockerClient dockerClient) {
...
   public static HashicorpNode createAndStartHashicorp(
       final DockerClient dockerClient, final boolean withTls) {
-    final HashicorpNode hashicorpNode =
-        withTls
-            ? new HashicorpNode(dockerClient, HashicorpVaultDockerCertificate.create())
-            : new HashicorpNode(dockerClient);
-    hashicorpNode.start();
-    return hashicorpNode;
+    try {
+      final SelfSignedCertificate selfSignedCertificate = SelfSignedCertificate.generate();
```
usmansaleem: nit: there is no point in generate if withTls is false. May I suggest:
~~~
      final HashicorpNode hashicorpNode =
          withTls
              ? new HashicorpNode(dockerClient, SelfSignedCertificate.generate())
              : new HashicorpNode(dockerClient);
~~~

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultDockerCertificate.java
```diff
@@ -53,16 +54,19 @@ private HashicorpVaultDockerCertificate(
...
-  public static HashicorpVaultDockerCertificate create() {
+  public static HashicorpVaultDockerCertificate create(
+      final SelfSignedCertificate selfSignedCertificate) {
     try {
-      final SelfSignedCertificate selfSignedCertificate = SelfSignedCertificate.generate();
       final Path certificateDirectory = createDestinationCertificateDirectory();
-      final Path tlsCertificate =
-          copyCertificate(selfSignedCertificate.certificatePath(), certificateDirectory);
-      final Path tlsPrivateKey =
-          copyCertificate(selfSignedCertificate.privateKeyPath(), certificateDirectory);
```
usmansaleem: Just a side note, I think docker-java API also provide support for "copyFileIntoContainer". Not as part of this PR, but in future we might be able to get away with all this fidling with mounting files if we copy the generated certificate into the container before starting hashicorp.

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpNode.java
rain-on: nice.

PR #9, opened 2020-03-13
Extract Vault commands to separate class

rain-on:


acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultDocker.java
```diff
@@ -60,11 +62,14 @@
...
   private static final int DEFAULT_VAULT_PORT = 8200;
 
-  private static final String SECRET_PATH = "/secret";
+  private static final String VAULT_ROOT_PATH = "/secret";
   private static final String SIGNING_KEY_RESOURCE = "/signingKeyResource";
-  private static final String VAULT_PUT_RESOURCE = SECRET_PATH + SIGNING_KEY_RESOURCE;
+  private static final String VAULT_PUT_RESOURCE = VAULT_ROOT_PATH + SIGNING_KEY_RESOURCE;
   private static final String VAULT_SIGNING_KEY_PATH =
-      "/v1" + SECRET_PATH + "/data" + SIGNING_KEY_RESOURCE;
+      "/v1" + VAULT_ROOT_PATH + "/data" + SIGNING_KEY_RESOURCE;
```
usmansaleem: should we ask user complete url path? i.e. `/v1/secret/data/MySecretKey`?

usmansaleem: instead of assuming that hashicorp will always run on `/v1` and that the admin has enabled `/secret`?

rain-on: Our production code assumes the user will input the FULL Http path to the key, rather than just "mysecretkey". 
Otherwise there are simply too many things to insert to the commandline eg:
* Vault Root Path
* Http API version
All of that information is effectively encoded in the key path.

usmansaleem: 👍

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultCommands.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.signers.dsl.hashicorp;
+
+import java.util.Map;
+
+// This assumes there is a Vault Server running at teh vault URL (for AT, typically in a docker)
```
jframe: nit: \teh\the

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultCommands.java
```diff
@@ -0,0 +1,64 @@
...
+public class HashicorpVaultCommands {
+
+  private final String vaultUrl;
+
+  public HashicorpVaultCommands(final String vaultUrl) {
+    this.vaultUrl = vaultUrl;
+  }
+
+  public String[] vaultStatusCommand() {
```
jframe: nit: no need to prefix all these commands with vault

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultCommands.java
```diff
@@ -0,0 +1,64 @@
...
+    };
+  }
+
+  public String[] unsealVault(final String unsealKey) {
+    return new String[] {
+      "vault", "operator", "unseal", "-address=" + vaultUrl, "-format=json", unsealKey
+    };
+  }
+
+  public String[] vaultPutSecretCommand(final Map.Entry<String, String> entry, final String path) {
```
jframe: why not just pass in the key and value as seperate params? seems odd to pass in an entry type

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultCommands.java
rain-on: true

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultCommands.java
rain-on: I guess in my head it was a kv store - so an entry made sense.
Also - I think we want to pass a whole Map into the layer above this, and that layer can have a simple loop "foreach entry - insert(entry, path)"

PR #12, opened 2020-03-13
Allow secrets to be added to a Hashicorp vault (AT DSL)

rain-on:


acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpNode.java
```diff
@@ -72,15 +73,18 @@ public String getHost() {
...
 
   public Optional<SelfSignedCertificate> getServerCertificate() {
     return serverTlsCertificate;
   }
+
+  /* Note: Path should be the "subpath" of the secret - not the full HTTP path.
+  The full HTTP Path will be returned.
+   */
+  public String addSecretsToVault(final Map<String, String> entries, final String path) {
+    return hashicorpVaultDocker.addSecretsToVault(entries, path);
```
usmansaleem: does it return the full URL? like `http://....//`?

usmansaleem: may be return a constructed URL? The method name is not very helpful :)

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultDocker.java
```diff
@@ -177,20 +165,32 @@ private void postStartup() {
     hashicorpRootToken = hashicorpVaultTokens.getRootToken();
   }
 
-  private void createSecretKey() {
+  public String addSecretsToVault(final Map<String, String> entries, final String path) {
```
usmansaleem: The method is doing two things, isn't it :).

acceptance-tests/src/test/java/tech/pegasys/signers/keystorage/hashicorp/tests/HashicorpVaultAccessAcceptanceTest.java
```diff
@@ -73,13 +76,17 @@ void cleanup() {
   void keyCanBeExtractedFromVault() throws IOException {
     hashicorpNode = HashicorpNode.createAndStartHashicorp(docker, false);
 
+    final String hashicorpSecretHttpPath =
+        hashicorpNode.addSecretsToVault(
+            Collections.singletonMap(SECRET_KEY, SECRET_VALUE), "acceptanceTestSecret");
+
```
usmansaleem: nit: We should use the utlity method directly here to construct the URL. The intent of "addSecretsToVault" should either be successful or failed.

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpVaultDocker.java
rain-on: as in, inserting the key, and returning the path?

usmansaleem: yeah ... feels a bit off tbh (or may be I am overthinking). You already have a separate method to construct the "path" from "/secret". Call that whereever the "path" is required and this method can remain void? not-a-show-stopper. This is just nit-picking.

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpNode.java
rain-on: now returns void - just adds a secret - its up the to test to construct the actual path (via helper function)

acceptance-tests/src/test/java/tech/pegasys/signers/dsl/hashicorp/HashicorpNode.java
```diff
@@ -72,15 +73,25 @@ public String getHost() {
...
   }
+
+  /* Note: Path should be the "subpath" of the secret - not the full HTTP path.
+  The full HTTP Path will be returned.
+   */
+  public void addSecretsToVault(final Map<String, String> entries, final String path) {
+    hashicorpVaultDocker.addSecretsToVault(entries, path);
+  }
+
+  public String getHttpApiPathForSecret(final String secretPath) {
```
usmansaleem: nice

PR #14, opened 2020-03-16
Refactor naming used for jars and publishing

rain-on:


build.gradle
```diff
@@ -35,7 +35,7 @@ if (!JavaVersion.current().java11Compatible) {
 def bintrayUser = project.hasProperty('bintrayUser') ? project.property('bintrayUser') : System.getenv('BINTRAY_USER')
 def bintrayKey = project.hasProperty('bintrayApiKey') ? project.property('bintrayApiKey') : System.getenv('BINTRAY_KEY')
 def bintrayPackage = bintray.pkg {
-  repo = 'pegasys-repo'
+  repo = 'test-repo'
```
jframe: This should stay as 'pegasys-repo'

build.gradle
```diff
@@ -240,6 +240,12 @@ task deploy() {}
 
 subprojects {
 
+  build {
```
jframe: Is this this needed? Looks like it was just for testing

build.gradle
rain-on: agreed

build.gradle
rain-on: agreed

PR #16, opened 2020-03-25
Remove teku BLS dependency

usmansaleem:
--Pass public key to encrypt method rather than calculating it from private key
 --Change version to 0.0.2-SNAPSHOT as this is breaking change with 0.0.1-SNAPSHOT
 -- Fix errorprone checks to allow build with JDK12+

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,2 @@
 org.gradle.jvmargs=-Xmx1g
-version=0.0.1-SNAPSHOT
+version=0.0.2-SNAPSHOT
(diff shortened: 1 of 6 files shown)


gradle.properties
```diff
@@ -1,2 +1,2 @@
 org.gradle.jvmargs=-Xmx1g
-version=0.0.1-SNAPSHOT
+version=0.0.2-SNAPSHOT
```
jframe: don't think we really need to update the version, it's snapshot not expected to be stable

PR #18, opened 2020-03-25
Setting up snapshot version

usmansaleem:
-- 1.1.0-SNAPSHOT


gradle.properties
```diff
@@ -1,2 +1,2 @@
 org.gradle.jvmargs=-Xmx1g
-version=1.0.0
+version=1.1.0-SNAPSHOT
```
jframe: 1.0.1-SNAPSHOT instead?

usmansaleem: done

PR #23, opened 2020-05-29
Add secp256k1 signing

jframe:
Adds secp256k1 signing taken from EthSigner. This includes signing implementations for azure, file and hashicorp.


settings.gradle
```diff
@@ -15,3 +15,9 @@ rootProject.name='signers'
 include 'bls-keystore'
 include 'keystorage:hashicorp'
 include 'acceptance-tests'
+include 'signing:secp-signer-api'
```
rain-on, line 18: wondering if this should be signing:secp-signer:api (rather than -api)?

jframe: that's nicer. done

jframe: gone a bit further and also dropped -signer from the module names. so now it signing:secp:api, signing:secp-azure etc.

PR #26, opened 2020-06-03
Collapse signing modules

jframe:
Remove the signing submodules as they didn't offer much value. Most modules only a few classes. This simplifies maintenance and makes it easier to import the signers.


signing/secp256k1/impl/build.gradle
```diff
@@ -25,26 +26,36 @@ jar {
   }
 }
 
+testFixturesJar {
+  archiveBaseName = calculateJarName(project)
```
rain-on, line 30: does this create the right named testfixture jar?

jframe: I copied this from what the hashicorp module was doing so it should be. I'll try and confirm. It is working correctly if I publish to maven locally

PR #28, opened 2020-06-09
Secp256k1 signing ATs

jframe:
Add acceptance tests for secp256k1 signing. These tests are ported from EthSigner.


acceptance-tests/src/test/java/tech/pegasys/signers/secp256k1/tests/multikey/HashicorpHelpers.java
```diff
@@ -0,0 +1,36 @@
...
+
+import tech.pegasys.signers.hashicorp.dsl.HashicorpNode;
+import tech.pegasys.signers.secp256k1.HashicorpSigningParams;
+
+import com.github.dockerjava.api.DockerClient;
+
+public class HashicorpHelpers {
+
+  public static final String secretPath = "acceptanceTestSecretPath";
+  public static final String secretName = "value"; // this is the required default usedin EthSigner
```
rain-on, line 25: nit: bad comment - does "value" still need to be used? If so - why?

jframe: No it doesn't have to be value at all. Signers doesn't care. I'll remove the comment

PR #39, opened 2020-07-21
ADHARA: HSM Extensions

mkrielza:


gradle/check-licenses.gradle
```diff
@@ -46,7 +46,8 @@ ext.acceptedLicenses = [
         'CC0 1.0 Universal License',
         'Common Development and Distribution License 1.0',
         'Unicode/ICU License',
-        'CC0'
+        'CC0',
+        'IAIK of Graz University of Technology License'
```
rain-on, line 50: need to investigate if we're happy with Ethsigner using this license - if not, maybe we can look at creating "plugins" for signers - such that this can be dynamically loaded, and its not longer an issue for Ethsigner.

mkrielza:
Closing PR as this is badly outdated after refactor.

PR #40, opened 2020-07-22
Allow PublicKey to be used as Identifier for a TransactionSigner

rain-on:
The TransactionSigner now can be queried for its PublicKey - which is a more generic identifier than Ethereum Address.
This in turn allows the TransactionSigner concept to be used in non-Ethereum applications.


signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/TransactionSigner.java
```diff
@@ -12,9 +12,12 @@
...
-
-  Signature sign(final byte[] data);
+/**
+ * This interface, and the "getAddress" API is to be removed, going forward Signers should implement
+ * the "Signer" interface directly.
+ */
+@Deprecated
+public interface TransactionSigner extends Signer {
 
   String getAddress();
```
jframe, line 22: Why the nested hierarchy? Could we just add the getPublicKey to the existing TransactionSigner and mark the getAddress as deprecated?

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/PublicKey.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.signers.secp256k1.api;
+
+public interface PublicKey {
+
+  byte[] getValue();
```
jframe, line 17: Bytes instead byte[] would be more convenient. It does however add a dependency on Tuweni but that might not be so bad

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SingleTransactionSignerProvider.java
```diff
@@ -45,4 +54,13 @@ public SingleTransactionSignerProvider(final TransactionSigner signer) {
       return Collections.emptySet();
     }
   }
+
+  @Override
+  public Set<PublicKey> availablePublicKeys() {
+    if (signer.getPublicKey() != null) {
```
jframe, line 60: nit: ternary would be simpler

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/PublicKeyImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import tech.pegasys.signers.secp256k1.api.PublicKey;
+
+import java.math.BigInteger;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.web3j.utils.Numeric;
+
+public class PublicKeyImpl implements PublicKey {
+
+  static final int PUBLIC_KEY_SIZE = 64;
```
jframe, line 24: nit: this can be private I think

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
```diff
@@ -50,15 +50,15 @@ public SigningMetadataTomlConfigLoader(final Path rootDirectory) {
     this.tomlConfigsDirectory = rootDirectory;
   }
 
-  Optional<SigningMetadataFile> loadMetadataForAddress(final String address) {
+  Optional<SigningMetadataFile> loadMetadata(final String identifier) {
```
jframe, line 53: Why the rename if this is still specific to Ethereum addresses?

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/PublicKey.java
rain-on: API literally has 0 dependencies - I would MUCH prefer to use Bytes, but thought the dependency bit was nicer.

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SingleTransactionSignerProvider.java
rain-on: done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/PublicKeyImpl.java
rain-on: done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
rain-on: its not longer specific - this gets used for Addresses and public keys - its now just a filename

jframe: The normalizeAddress is calls below through me off. Do we still need to do the normalizeAddress in this function?

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
```diff
@@ -45,20 +46,30 @@
...
+
+  Optional<SigningMetadataFile> loadMetadata(final String identifier) {
     final List<SigningMetadataFile> matchingMetadata =
         loadAvailableSigningMetadataTomlConfigs().stream()
             .filter(
-                toml -> toml.getBaseFilename().toLowerCase().endsWith(normalizeAddress(address)))
+                toml ->
+                    toml.getBaseFilename()
+                        .toLowerCase()
+                        .endsWith(identityFilenameTranslator.apply(identifier).toLowerCase()))
```
jframe, line 68: can you add a test to show that the translator is working?

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
```diff
@@ -45,20 +46,30 @@
   private static final String GLOB_CONFIG_MATCHER = "**" + CONFIG_FILE_EXTENSION;
 
   private final Path tomlConfigsDirectory;
+  private final Function<String, String> identityFilenameTranslator;
 
-  public SigningMetadataTomlConfigLoader(final Path rootDirectory) {
+  public SigningMetadataTomlConfigLoader(
+      final Path rootDirectory, final Function<String, String> identityFilenameTranslator) {
```
jframe, line 52: nit: Perhaps an interface so would be nicer than Function<String, String> so we can give this type a name

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeyTransactionSignerProvider.java
```diff
@@ -39,72 +43,92 @@
...
+              try {
+                if ((signer != null)
+                    && configFileSelector
+                        .getSpecificConfigFileFilter(signer.getPublicKey())
+                        .accept(Path.of(metadataFile.getFilename()))) {
+                  return signer;
+                }
+                return null;
+              } catch (final IOException e) {
+                return null;
```
usmansaleem, line 98: a warn log might be useful to identify why IOException is raised (security issue, disk full etc.)

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public boolean matchesPublicKey(final PublicKey publicKey, final Path entry) throws IOException {
+    String expectedFilename = publicKey.toString();
+    if (expectedFilename.startsWith("0x")) {
+      expectedFilename = expectedFilename.substring(2);
+    }
+
+    final List<String> tokens =
+        Splitter.onPattern("\\.(?=[^\\.]+$)").splitToList(entry.getFileName().toString());
```
usmansaleem, line 52: why are you using regexp? Is it because you want to compare the file name without case sensitivity but the extension must be lowercase `.toml`? (or extension can be case insensitive ?)

`entry.getFileName().toString().equalsIgnoreCase(expectedFileName + ".toml")` might work similarly.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
```diff
@@ -67,18 +62,19 @@ public SigningMetadataTomlConfigLoader(final Path rootDirectory) {
     }
   }
 
-  Collection<SigningMetadataFile> loadAvailableSigningMetadataTomlConfigs() {
-    final Collection<SigningMetadataFile> metadataConfigs = new HashSet<>();
+  ImmutableList<SigningMetadataFile> loadAvailableSigningMetadataTomlConfigs(
```
usmansaleem, line 65: Why return dependency on ImmutableList? You can declare `List` instead (assuming ImmutableList extends/implements List).

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/metadata/SigningMetadataFile.java
```diff
@@ -20,20 +20,12 @@
   protected String baseFilename;
 
   public SigningMetadataFile(final String filename) {
-    this.baseFilename = getFilenameWithoutExtension(filename);
+    this.baseFilename = filename;
```
jframe, line 23: can rename this to filename

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/FileSelector.java
```diff
@@ -0,0 +1,28 @@
...
+import java.nio.file.Path;
+
+/*
+A set of rules regarding how key files are to be selected.
+Should be overridden by the calling application to define their naming
+convention.
+ */
+public interface FileSelector<T> {
+
+  Filter<Path> getCollectiveFilter();
```
jframe, line 25: getAllConfigFilesFilter? getCollective.. just sounds a bit odd to me

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeyTransactionSignerProvider.java
rain-on: done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
rain-on: Thinking this class can just be removed - its only used in a test ...

rain-on: OR do we want to leave it in as a sample?
The regex is nicer than the suggested - as it lets people put <publickey>.myfile.toml for example.
This is not actually used in production code...

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
rain-on: true. Done.

acceptance-tests/src/test/java/tech/pegasys/signers/secp256k1/tests/multikey/FileBasedTomlLoadingAcceptanceTest.java
```diff
@@ -22,11 +22,14 @@
...
+import org.apache.tuweni.bytes.Bytes;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
 class FileBasedTomlLoadingAcceptanceTest extends MultiKeyAcceptanceTestBase {
 
+  static final String FILE_PUBLIC_KEY =
+      "0xb0de161ce49581aa75f25852fb1bb882c8921ff5f6eaca9329f5c5f34966085a704f0a555c148359ebaffbe0d92712386890606a5ac8e54563db8279f2120f5d";
   static final String FILENAME = "a01f618424b0113a9cebdc6cb66ca5b48e9120c5";
   static final String FILE_ETHEREUM_ADDRESS = "0x" + FILENAME;
```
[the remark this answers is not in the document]
rain-on: cleaned up - and this needed to stay

gradle.properties
```diff
@@ -1,4 +1,4 @@
 org.gradle.jvmargs=-Xmx1g
-version=1.0.3-SNAPSHOT
+version=1.1.3-SNAPSHOT
```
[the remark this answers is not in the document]
rain-on: moved back to 1.0.3-SNAPSHOT

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeyTransactionSignerProvider.java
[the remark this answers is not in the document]
rain-on: The Filter forces there to be an "Throws IOException" - so it needs to be handled, but if it fails, the right answer for this function is to return null rather than letting the exception escape.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
```diff
@@ -0,0 +1,65 @@
...
+    return this::hasExpectedFileExtension;
+  }
+
+  @Override
+  public Filter<Path> getSpecificConfigFileFilter(final PublicKey selectionCriterion) {
+    return entry -> matchesPublicKey(selectionCriterion, entry);
+  }
+
+  public boolean matchesPublicKey(final PublicKey publicKey, final Path entry) throws IOException {
+    String expectedFilename = publicKey.toString();
```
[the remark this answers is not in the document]
rain-on: can't be final, as it gets modified

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/SigningMetadataTomlConfigLoader.java
```diff
@@ -67,18 +62,19 @@ public SigningMetadataTomlConfigLoader(final Path rootDirectory) {
...
+        Files.newDirectoryStream(tomlConfigsDirectory, configFileSelector)) {
       for (final Path file : directoryStream) {
         getMetadataInfo(file).ifPresent(metadataConfigs::add);
       }
-      return metadataConfigs;
+      return ImmutableList.copyOf(metadataConfigs);
     } catch (final IOException e) {
       LOG.warn("Error searching for signing metadata TOML files", e);
-      return Collections.emptySet();
+      return ImmutableList.of();
```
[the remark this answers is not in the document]
rain-on: done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/metadata/SigningMetadataFile.java
rain-on: done

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/FileSelector.java
rain-on: done

acceptance-tests/src/test/java/tech/pegasys/signers/secp256k1/tests/multikey/signing/MultiKeyTransactionSigningAcceptanceTestBase.java
```diff
@@ -33,13 +34,15 @@
 
 public class MultiKeyTransactionSigningAcceptanceTestBase extends MultiKeyAcceptanceTestBase {
 
-  private static final String ADDRESS = "fe3b557e8fb62b89f4916b721be55ceb828dbd73";
   private static final byte[] DATA_TO_SIGN = "42".getBytes(UTF_8);
   private static final String PRIVATE_KEY =
       "8f2a55949038a9610f50fb23b5883af3b4ecb3c3bb792cbcefbd1542c692be63";
+  static final String PUBLIC_KEY_HEX_STRING =
+      "0x9b02f8a5fddd222ade4ea4528faefc399623af3f736be3c44f03e2df22fb792f3931a4d9573d333ca74343305762a753388c3422a86d98b713fc91c1ea04842";
```
usmansaleem, line 41: I noticed that "FILENAME" in other test classes start with `09b02...`, is that leading 0 intentional (i.e. Bytes' class toHexString results in that?)

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SingleSignerProvider.java
```diff
@@ -12,15 +12,16 @@
...
+public class SingleSignerProvider implements SignerProvider {
 
-  private final TransactionSigner signer;
+  private final Signer signer;
 
-  public SingleTransactionSignerProvider(final TransactionSigner signer) {
+  public SingleSignerProvider(final Signer signer) {
     if (signer == null) {
       throw new IllegalArgumentException(
           "SingleTransactionSignerFactory requires a non-null TransactionSigner");
```
usmansaleem, line 27: exception message still referencing to "TransactionSigner". (might perform a full text search)

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SingleSignerProvider.java
```diff
@@ -29,20 +30,17 @@ public SingleTransactionSignerProvider(final TransactionSigner signer) {
   }
 
   @Override
-  public Optional<TransactionSigner> getSigner(final String address) {
-    if (signer.getAddress() != null && signer.getAddress().equalsIgnoreCase(address)) {
+  public Optional<Signer> getSigner(final PublicKey publicKey) {
+    if ((signer.getPublicKey() != null)
```
usmansaleem, line 34: `signer.getPublicKey() != null` is performed , but not the method argument publicKey (which might result in NPE).

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
```diff
@@ -33,75 +36,95 @@
...
+              try {
+                if ((signer != null)
+                    && configFileSelector
+                        .getSpecificConfigFileFilter(signer.getPublicKey())
+                        .accept(Path.of(metadataFile.getFilename()))) {
+                  return signer;
+                }
+                return null;
+              } catch (final IOException e) {
+                LOG.warn("IO Exception raised while loading {}", metadataFile.getFilename());
```
usmansaleem, line 97: may I suggest changing the warning to:

`LOG.warn("IO Exception raised while loading {}:{}", metadataFile.getFilename(), e.getMessage(),e);

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
```diff
@@ -33,75 +36,95 @@
...
   }
 
   @Override
-  public TransactionSigner createSigner(final AzureSigningMetadataFile metadataFile) {
+  public Signer createSigner(final AzureSigningMetadataFile metadataFile) {
     try {
       return azureFactory.createSigner(metadataFile.getConfig());
     } catch (final TransactionSignerInitializationException e) {
-      LOG.error("Failed to construct Azure signer from " + metadataFile.getBaseFilename());
+      LOG.error("Failed to construct Azure signer from " + metadataFile.getFilename());
```
usmansaleem, line 111: add `e` as argument to error to get the whole stacktrace (or print actual error message from `TransactionSignerInitializationException` at least?).

usmansaleem: `TransactionSignerInitializationException` should be renamed to `SignerInitializationException` ??

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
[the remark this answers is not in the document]
usmansaleem:
~~~
final String publicKeyStr = publicKey.toString();
final String expectedFileName = publicKeyStr.startsWith("0x") ? publicKeyStr.substring(2) : publicKeyStr;
~~~
would avoid declaring non-final variable.

acceptance-tests/src/test/java/tech/pegasys/signers/secp256k1/tests/multikey/signing/MultiKeyTransactionSigningAcceptanceTestBase.java
rain-on: leading 0 is required ... but this tests all seem the same, but different, so may spend some time looking into it

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SingleSignerProvider.java
rain-on: fixed.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
rain-on: you may :) Lovely.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
rain-on: Renamed exception, I'm wondering if this is one of the places we deliberately didn't want the full stack trace (due to noise) - that maybe(?) this message provides sufficient information to resolve the issue.

acceptance-tests/src/testFixtures/java/tech.pegasys.signers.secp256k1/DefaultFileSelector.java
```diff
@@ -0,0 +1,35 @@
...
+  @Override
+  public Filter<Path> getAllConfigFilesFilter() {
+    return entry -> entry.getFileName().toString().endsWith("toml");
+  }
+
+  @Override
+  public Filter<Path> getSpecificConfigFileFilter(final PublicKey publicKey) {
+    return entry -> {
+      final String filename = publicKey.toString().substring(2); // remove 0x prefix
+      return entry.getFileName().toString().endsWith(filename + ".toml");
```
usmansaleem, line 32: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#getFileName() ... if you use `endsWith`, there is a chance you will pick up multiple files which has different prefix but ends with `publickey.toml`. If thats the intention its alright, otherwise, `equals` might be appropriate.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/metadata/HashicorpSigningMetadataFile.java
```diff
@@ -31,7 +31,7 @@ public HashicorpKeyConfig getConfig() {
   }
 
   @Override
-  public TransactionSigner createSigner(MultiSignerFactory factory) {
+  public Signer createSigner(MultiSignerFactory factory) {
```
usmansaleem, line 34: nit: `final MultiSignerFactory factory`

signing/secp256k1/impl/src/testFixtures/java/tech/pegasys/signers/secp256k1/common/AddressPostFixFilter.java
```diff
@@ -0,0 +1,48 @@
...
+    if (tokens.size() < 2) {
+      return false;
+    }
+
+    return tokens.get(0).toLowerCase().equals(addressToMatch.toLowerCase())
+        && hasExpectedFileExtension(entry);
+  }
+
+  private boolean hasExpectedFileExtension(final Path entry) {
+    return entry.getFileName().toString().toLowerCase().endsWith(fileExtension.toLowerCase());
```
usmansaleem, line 46: `fileExtension` is already lowercase.

signing/secp256k1/impl/src/testFixtures/java/tech/pegasys/signers/secp256k1/common/AddressPostFixFilter.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Path;
+import java.util.List;
+
+import com.google.common.base.Splitter;
+
+public class AddressPostFixFilter implements DirectoryStream.Filter<Path> {
+
+  private static final String fileExtension = "toml";
```
usmansaleem, line 24: nit: since its a constant, may be use FILE_EXTENSION?

acceptance-tests/src/testFixtures/java/tech.pegasys.signers.secp256k1/DefaultFileSelector.java
rain-on: Yup - agreed - I think this filter should only return a single file, and it should be the simplest thing it can do - ie. <publickey>.toml.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/metadata/HashicorpSigningMetadataFile.java
rain-on: done

signing/secp256k1/impl/src/testFixtures/java/tech/pegasys/signers/secp256k1/common/AddressPostFixFilter.java
rain-on: done

signing/secp256k1/impl/src/testFixtures/java/tech/pegasys/signers/secp256k1/common/AddressPostFixFilter.java
rain-on: done

acceptance-tests/src/testFixtures/java/tech.pegasys.signers.secp256k1/DefaultFileSelector.java
```diff
@@ -0,0 +1,35 @@
...
+  @Override
+  public Filter<Path> getAllConfigFilesFilter() {
+    return entry -> entry.getFileName().toString().endsWith("toml");
+  }
+
+  @Override
+  public Filter<Path> getSpecificConfigFileFilter(final PublicKey publicKey) {
+    return entry -> {
+      final String filename = publicKey.toString().substring(2); // remove 0x prefix
+      return entry.getFileName().toString().equals(filename + ".toml");
```
jframe, line 32: Do we also want to match files with the 0x prefix? I think this is the current behaviour.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/PublicKeyImpl.java
```diff
@@ -0,0 +1,45 @@
...
+  @Override
+  public byte[] getValue() {
+    return value.toArrayUnsafe();
+  }
+
+  @Override
+  public String toString() {
+    return value.toHexString();
+  }
+}
```
jframe, line 45: Thinking this should implement equals as well

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
jframe, line 97: can you split this out into a seperate function

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
```diff
@@ -0,0 +1,65 @@
...
+  public Filter<Path> getAllConfigFilesFilter() {
+    return this::hasExpectedFileExtension;
+  }
+
+  @Override
+  public Filter<Path> getSpecificConfigFileFilter(final PublicKey selectionCriterion) {
+    return entry -> matchesPublicKey(selectionCriterion, entry);
+  }
+
+  public boolean matchesPublicKey(final PublicKey publicKey, final Path entry) throws IOException {
```
jframe, line 45: Doesn't look like the throws for IOException is needed

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/filebased/FileBasedTransactionSignerTest.java
```diff
@@ -51,11 +51,10 @@ public void success() throws IOException {
     final File keyFile = new File(fileName);
     final File pwdFile = createFile(MY_PASSWORD);
 
-    final TransactionSigner signer =
-        FileBasedSignerFactory.createSigner(keyFile.toPath(), pwdFile.toPath());
+    final Signer signer = FileBasedSignerFactory.createSigner(keyFile.toPath(), pwdFile.toPath());
 
     assertThat(signer).isNotNull();
-    assertThat(signer.getAddress()).isNotEmpty();
+    assertThat(signer.getPublicKey()).isNotNull();
```
jframe, line 57: This seems like a weak test. But maybe it's ok because we test that public key is what we expect elsewhere?

signing/secp256k1/impl/src/testFixtures/java/tech/pegasys/signers/secp256k1/common/AddressPostFixFilter.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package tech.pegasys.signers.secp256k1.common;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Path;
+
+import com.google.common.io.Files;
+
+public class AddressPostFixFilter implements DirectoryStream.Filter<Path> {
```
jframe, line 21: Didn't look like this was being used? Is it for tests will be needed in EthSigner?

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
jframe: Don't see any need to leave this as an example. The test implementation makes it clear enough on how write a FileSelector

acceptance-tests/src/testFixtures/java/tech.pegasys.signers.secp256k1/DefaultFileSelector.java
rain-on: This isn't the definitive solution - its a basic Filter that people can use if they do so wish, so didn't want to go to far with what it was meant to do ...

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/PublicKeyImpl.java
rain-on: done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
rain-on: done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/PublicKeyFileSelector.java
rain-on: agreed - but the fileFilter run as part of the getSpecificConfiFileFilter _can_ throw an IOException as part of its inherited function ... how do we avoid that?

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/filebased/FileBasedTransactionSignerTest.java
rain-on: I suspect the problem is that the WalletUtils doesn't tell you what the publickey is - this just proves that one was created :/ Yeah, a bit weak... but maybe not worth chasing?

signing/secp256k1/impl/src/testFixtures/java/tech/pegasys/signers/secp256k1/common/AddressPostFixFilter.java
rain-on: removed

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/PublicKeyImpl.java
```diff
@@ -0,0 +1,63 @@
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    final PublicKeyImpl other = (PublicKeyImpl) o;
+    return other.value.compareTo(value) == 0;
```
usmansaleem, line 56: You could have used `Objects.equals(other.value, value)` to match `Objects.hash` but is just nit pick.

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/filebased/FileBasedTransactionSignerTest.java
jframe: no don't think it is

PR #43, opened 2020-07-29
Validates the signer created matches requested public key

rain-on:


signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
```diff
@@ -73,9 +74,21 @@ public MultiKeySignerProvider(
...
   public Optional<Signer> getSigner(final PublicKey publicKey) {
-    return signingMetadataTomlConfigLoader
-        .loadMetadata(configFileSelector.getSpecificConfigFileFilter(publicKey))
-        .map(metadataFile -> metadataFile.createSigner(this));
+    final Optional<Signer> signer =
+        signingMetadataTomlConfigLoader
+            .loadMetadata(configFileSelector.getSpecificConfigFileFilter(publicKey))
+            .map(metadataFile -> metadataFile.createSigner(this));
+    if (signer.isPresent()) {
+      if (Arrays.equals(signer.get().getPublicKey().getValue(), publicKey.getValue())) {
```
jframe, line 82: This shouldn't be necessary to use Arrays.equals as the PublicKey type implements equals

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProviderTest.java
```diff
@@ -132,4 +132,22 @@ void signerIsLoadedSuccessfullyWhenAddressHasCaseMismatchToFilename() throws URI
     assertThat(signer).isNotNull();
     assertThat(signer.getPublicKey().toString()).isEqualTo("0x" + LOWER_CASE_PUBLIC_KEY);
   }
+
+  @Test
+  void signerIsNotReturnedIfFileContainsPublicKeyDifferentToRequest() throws URISyntaxException {
+
+    final FileBasedSigningMetadataFile capitalisedMetadata =
+        new FileBasedSigningMetadataFile(
+            LOWERCASE_ADDRESS.toUpperCase() + ".toml",
```
jframe, line 141: Does the address need to be uppercase? Didn't think this would matter for this test

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
rain-on: The _base_ publicKey type doesn't offer equals - only the PublickKeyImpl - which is not _necessarily_ being used - thus you need to use the interface being offered by the PublicKey type, no?

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProviderTest.java
rain-on: correct, sign is irrelevant for this test.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
jframe: Since there a interface yes, we can't guarantee that is implemented

PR #47, opened 2020-08-05
Use ecpoint for publickey

jframe:
Change public key type so that it now holds a ECPoint instead of the byte value. This allows for using non-ethereum Public keys that might be serialised to bytes or string differently.


acceptance-tests/src/test/java/tech/pegasys/signers/secp256k1/tests/multikey/AzureBasedTomlLoadingAcceptanceTest.java
```diff
@@ -50,9 +51,7 @@ void azureSignersAreCreatedAndExpectedAddressIsReported(@TempDir Path tomlDirect
 
     setup(tomlDirectory);
 
-    assertThat(
-            signerProvider.availablePublicKeys().stream()
-                .map(pk -> Bytes.wrap(pk.getValue()).toUnprefixedHexString()))
-        .containsOnly(PUBLIC_KEY_HEX_STRING);
+    assertThat(signerProvider.availablePublicKeys().stream().map(PublicKey::toEthHexString))
```
rain-on, line 54: _should_ theEthHexString be _on_ the public key, or a helperfunction? I _prefer_ static helper-function, as it keeps "PublicKey" really clean and single purpose.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.tuweni.bytes.Bytes;
+import org.apache.tuweni.bytes.Bytes32;
+import org.web3j.utils.Numeric;
+
+public class EthPublicKeyUtils {
+  private static final int PUBLIC_KEY_SIZE = 64;
+
+  public static ECPublicKey createPublicKey(final ECPoint publicPoint) {
+    try {
+      AlgorithmParameters parameters = AlgorithmParameters.getInstance("EC");
```
rain-on, line 38: static class level variable?

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
```diff
@@ -0,0 +1,72 @@
...
+      ECPublicKeySpec pubSpec = new ECPublicKeySpec(publicPoint, ecParameters);
+      KeyFactory kf = KeyFactory.getInstance("EC");
+      return (ECPublicKey) kf.generatePublic(pubSpec);
+    } catch (NoSuchAlgorithmException | InvalidParameterSpecException | InvalidKeySpecException e) {
+      throw new IllegalStateException("Unable to create Ethereum public key", e);
+    }
+  }
+
+  public static ECPublicKey createPublicKey(final Bytes value) {
+    final Bytes x = value.slice(0, 32);
```
rain-on, line 50: add some validation that value is 64 bytes long, should it have a leading "4" or not? (assuming not)

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
```diff
@@ -0,0 +1,72 @@
...
+    final Bytes y = value.slice(32, 32);
+    final ECPoint ecPoint =
+        new ECPoint(Numeric.toBigInt(x.toArrayUnsafe()), Numeric.toBigInt(y.toArrayUnsafe()));
+    return createPublicKey(ecPoint);
+  }
+
+  public static ECPublicKey createPublicKey(final BigInteger value) {
+    final Bytes ethBytes = Bytes.wrap(Numeric.toBytesPadded(value, PUBLIC_KEY_SIZE));
```
rain-on, line 58: what if value is > 64 bytes?

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public static ECPublicKey createPublicKey(final Bytes value) {
+    final Bytes x = value.slice(0, 32);
+    final Bytes y = value.slice(32, 32);
+    final ECPoint ecPoint =
+        new ECPoint(Numeric.toBigInt(x.toArrayUnsafe()), Numeric.toBigInt(y.toArrayUnsafe()));
+    return createPublicKey(ecPoint);
+  }
+
+  public static ECPublicKey createPublicKey(final BigInteger value) {
```
rain-on, line 57: is there a symmetric call to this, i.e. ECPUblicKey --> BigInteger?

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtilsTest.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.signers.secp256k1;
+
+class EthPublicKeyUtilsTest {
+
+  //  @Test
```
rain-on, line 17: does this need to come back in? Or is it covered elsewhere (and thus should be removed?)

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
jframe: It can't be because it throws a AlgorithmNotFoundException

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
jframe: done. nope it shouldn't have a leading 4 as this for Ethereum public keys which leave that out.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
jframe: then Numeric throws a RuntimeException with message "Input is too large to put in byte array of size " + length

signing/secp256k1/impl/src/test/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtilsTest.java
jframe: yes it does. done

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/EthPublicKeyUtils.java
jframe: No not at the moment. I only added functions we currently need. Should we add this though?

rain-on: Sounds like a YAGNI - don't worry about it

PR #53, opened 2020-08-19
Allow CredentialSigner to take data-hashes

rain-on:
The CredentialSigner uses the underlying web3j implementation to perform signing - at the moment, web3j will always hash the supplied data prior to producing a signature.

This update allows the CredentialSigner to be created  with or without enabling said hashing - this in turn means users of the class can supply a pre-hashed data object for signing.


jframe (approved):
LGTM

PR #58, opened 2020-08-25
Signer verify

jframe:
Add verify operation to Signer. This has been implemented for both the CredentialSigner and the AzureKeyVaultSigner. This operation is needed for new API that is being added to Eth2Signer.


usmansaleem (approved):
Looks good to me.

rain-on (approved):
LGTM

jframe:
After further discussion, this isn't needed in the signer library we can do a EcRecover for the one place that requires this and avoid changes to the Signer API.

PR #59, opened 2020-08-31
Implement unicode password support from EIP2335

ajsutton:
EIP2335 has been updated to explicitly make passwords unicode and specify how to normalise them.  This implements that for signers.  Note that because control characters are now stripped from passwords as part of normalisation, there is no need for callers to strip line endings when reading files and the should just read the entire file content.

See https://github.com/ethereum/EIPs/pull/2749/files for spec changes.

### DIFF
--- a/bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/PasswordUtils.java
+++ b/bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/PasswordUtils.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2020 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/PasswordUtils.java
```diff
@@ -0,0 +1,36 @@
...
+    final String normalizedPassword = Normalizer.normalize(password, Form.NFKD);
+    final int[] filteredCodepoints =
+        normalizedPassword.chars().filter(c -> !isControlCode(c)).toArray();
+    final byte[] utf8Password =
+        new String(filteredCodepoints, 0, filteredCodepoints.length).getBytes(UTF_8);
+    return Bytes.wrap(utf8Password);
+  }
+
+  private static boolean isControlCode(final int c) {
+    return (0x00 <= c && c <= 0x1F) || (0x80 <= c && c <= 0x9F) || c == 0x7F;
```
rain-on, line 34: _could_ split this into sub functions to highlight the zones explicitly - but that's nit picking.

rain-on (approved):
LGTM

rain-on (commented):
LGTM - ship it.

PR #64, opened 2020-09-15
Azure secret list

rain-on:
Allows the AzureKeyVault to expose the names of all available keys, such that they can be iterated.

### DIFF
--- a/keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
+++ b/keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
@@ -12,7 +12,9 @@
  */
 package tech.pegasys.signers.azure;
 
+import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import com.azure.core.exception.ResourceNotFoundException;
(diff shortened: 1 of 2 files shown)


keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
```diff
@@ -78,4 +81,10 @@ public CryptographyClient fetchKey(final String keyName, final String keyVersion
   public static String constructAzureKeyVaultUrl(final String keyVaultName) {
     return String.format(AZURE_URL_PATTERN, keyVaultName);
   }
+
+  public List<String> getAvailableSecrets() {
```
usmansaleem, line 85: nit: should we return a Set instead?

usmansaleem (approved):
LGTM

keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
usmansaleem: also, should method name be `getAvailableSecretNames()` ?

PR #65, opened 2020-09-15
Allow raw signers via toml

rain-on:


signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
```diff
@@ -144,6 +146,17 @@ public Signer createSigner(final FileBasedSigningMetadataFile metadataFile) {
     }
   }
 
+  @Override
+  public Signer createSigner(final String privateKey) {
+    try {
+      final Credentials credentials = Credentials.create(privateKey);
+      return new CredentialSigner(credentials);
+    } catch (final Exception e) {
+      LOG.error("Unable to construct Raw signer from {}", privateKey);
```
usmansaleem, line 155: I think we should also append `e.getMessage()`.

usmansaleem (approved):
LGTM - minor suggestion around adding exception message to the log as well.

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
rain-on: Changed this to better reflect how all other signers are created (i.e. from a MetadataFile, rather than a raw string).

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
```diff
@@ -144,6 +147,17 @@ public Signer createSigner(final FileBasedSigningMetadataFile metadataFile) {
     }
   }
 
+  @Override
+  public Signer createSigner(final RawSigningMetadataFile metadataFile) {
+    try {
+      final Credentials credentials = Credentials.create(metadataFile.getPrivKey());
+      return new CredentialSigner(credentials);
+    } catch (final Exception e) {
+      LOG.error("Unable to construct Raw signer from " + metadataFile.getFilename());
```
jframe, line 156: nit: why use capital R for Raw? seems odd

jframe (approved):
LGTM

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/multikey/MultiKeySignerProvider.java
rain-on: changed to lower case

PR #68, opened 2020-09-16
YubiHSM2 Module

usmansaleem:
Provide support to read opaque data from YubiHSM2 using yubihsm-shell process.

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -68,5 +68,9 @@ dependencyManagement {
     dependency 'org.web3j:core:4.5.14'
 
     dependency 'org.zeroturnaround:zt-exec:1.11'
+
+    dependency 'info.picocli:picocli:4.5.1'
+
+    dependency 'tech.pegasys.teku.internal:bls:0.12.6'
   }
 }
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -68,9 +68,5 @@ dependencyManagement {
     dependency 'org.web3j:core:4.5.14'
 
     dependency 'org.zeroturnaround:zt-exec:1.11'
-
-    dependency 'info.picocli:picocli:4.5.1'
-
-    dependency 'tech.pegasys.teku.internal:bls:0.12.6'
   }
 }
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -25,6 +25,8 @@ dependencyManagement {
 
     dependency 'commons-cli:commons-cli:1.4'
 
+    dependency 'commons-io:commons-io:2.8.0'
+
     dependency 'io.rest-assured:rest-assured:4.2.0'
 
     dependency 'io.vertx:vertx-codegen:3.8.2'
@@ -34,6 +36,8 @@ dependencyManagement {
     dependency 'io.vertx:vertx-web:3.8.2'
 
     dependency 'javax.activation:activation:1.1.1'
+
+    dependency 'org.apache.commons:commons-lang3:3.11'
 
     dependency 'org.apache.logging.log4j:log4j-api:2.13.0'
     dependency 'org.apache.logging.log4j:log4j-core:2.13.0'
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -68,5 +68,7 @@ dependencyManagement {
     dependency 'org.web3j:core:4.5.14'
 
     dependency 'org.zeroturnaround:zt-exec:1.11'
+
+    dependency 'info.picocli:picocli:4.5.1'
   }
 }
--- a/keystorage/yubihsm2/build.gradle
+++ b/keystorage/yubihsm2/build.gradle
@@ -35,9 +35,9 @@ dependencies {
   testImplementation 'org.mockito:mockito-core'
   testImplementation 'org.apache.commons:commons-lang3'
   testImplementation 'com.google.guava:guava'
-  testImplementation 'info.picocli:picocli'
-  testImplementation 'tech.pegasys.teku.internal:bls'
-  testImplementation 'org.apache.tuweni:tuweni-bytes'
+
+  // required for resources/YubiShellSimulator.java
+  testRuntimeOnly 'info.picocli:picocli:4.5.1'
 
   errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/yubihsm2/build.gradle
+++ b/keystorage/yubihsm2/build.gradle
@@ -7,7 +7,6 @@
  */
 
 apply plugin: 'java-library'
-apply plugin: 'java-test-fixtures'
 
 jar {
   archiveBaseName = calculateJarName(project)
@@ -35,6 +34,10 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
   testImplementation 'org.mockito:mockito-core'
   testImplementation 'org.apache.commons:commons-lang3'
+  testImplementation 'com.google.guava:guava'
+  testImplementation 'info.picocli:picocli'
+  testImplementation 'tech.pegasys.teku.internal:bls'
+  testImplementation 'org.apache.tuweni:tuweni-bytes'
 
   errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/yubihsm2/build.gradle
+++ b/keystorage/yubihsm2/build.gradle
@@ -10,28 +10,28 @@ apply plugin: 'java-library'
 apply plugin: 'java-test-fixtures'
 
 jar {
-    archiveBaseName = calculateJarName(project)
-    manifest {
-        attributes(
-                'Specification-Title': archiveBaseName.get(),
-                'Specification-Version': rootProject.version,
-                'Implementation-Title': archiveBaseName.get(),
-                'Implementation-Version': calculateVersion()
-        )
-    }
+  archiveBaseName = calculateJarName(project)
+  manifest {
+    attributes(
+      'Specification-Title': archiveBaseName.get(),
+      'Specification-Version': rootProject.version,
+      'Implementation-Title': archiveBaseName.get(),
+      'Implementation-Version': calculateVersion()
+      )
+  }
 }
 
 dependencies {
-    implementation 'org.apache.logging.log4j:log4j-api'
+  implementation 'org.apache.logging.log4j:log4j-api'
 
-    runtimeOnly 'org.apache.logging.log4j:log4j-core'
-    runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+  runtimeOnly 'org.apache.logging.log4j:log4j-core'
+  runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
 
-    testImplementation 'org.assertj:assertj-core'
-    testImplementation 'org.junit.jupiter:junit-jupiter-engine'
-    testImplementation 'org.junit.jupiter:junit-jupiter-api'
-    testImplementation 'org.junit.jupiter:junit-jupiter-params'
-    testImplementation 'org.mockito:mockito-core'
+  testImplementation 'org.assertj:assertj-core'
+  testImplementation 'org.junit.jupiter:junit-jupiter-engine'
+  testImplementation 'org.junit.jupiter:junit-jupiter-api'
+  testImplementation 'org.junit.jupiter:junit-jupiter-params'
+  testImplementation 'org.mockito:mockito-core'
 
-    errorprone("com.google.errorprone:error_prone_core")
+  errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/yubihsm2/build.gradle
+++ b/keystorage/yubihsm2/build.gradle
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2020 ConsenSys AG.
+ *
+ * The source code is provided to licensees of PegaSys Plus for their convenience and internal
+ * business use. These files may not be copied, translated, and/or distributed without the express
+ * written permission of an authorized signatory for ConsenSys AG.
+ */
+
+apply plugin: 'java-library'
+apply plugin: 'java-test-fixtures'
+
+jar {
+    archiveBaseName = calculateJarName(project)
+    manifest {
+        attributes(
+                'Specification-Title': archiveBaseName.get(),
+                'Specification-Version': rootProject.version,
+                'Implementation-Title': archiveBaseName.get(),
+                'Implementation-Version': calculateVersion()
+        )
+    }
+}
+
+dependencies {
+    implementation 'org.apache.logging.log4j:log4j-api'
+
+    runtimeOnly 'org.apache.logging.log4j:log4j-core'
+    runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
+
+    testImplementation 'org.assertj:assertj-core'
+    testImplementation 'org.junit.jupiter:junit-jupiter-engine'
+    testImplementation 'org.junit.jupiter:junit-jupiter-api'
+    testImplementation 'org.junit.jupiter:junit-jupiter-params'
+    testImplementation 'org.mockito:mockito-core'
+
+    errorprone("com.google.errorprone:error_prone_core")
+}
--- a/keystorage/yubihsm2/build.gradle
+++ b/keystorage/yubihsm2/build.gradle
@@ -37,7 +37,7 @@ dependencies {
   testImplementation 'com.google.guava:guava'
 
   // required for resources/YubiShellSimulator.java
-  testRuntimeOnly 'info.picocli:picocli:4.5.1'
+  testRuntimeOnly 'info.picocli:picocli'
 
   errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/yubihsm2/build.gradle
+++ b/keystorage/yubihsm2/build.gradle
@@ -23,6 +23,8 @@ jar {
 
 dependencies {
   implementation 'org.apache.logging.log4j:log4j-api'
+  implementation 'org.zeroturnaround:zt-exec'
+  implementation 'commons-io:commons-io'
 
   runtimeOnly 'org.apache.logging.log4j:log4j-core'
   runtimeOnly 'org.apache.logging.log4j:log4j-slf4j-impl'
@@ -32,6 +34,7 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-api'
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
   testImplementation 'org.mockito:mockito-core'
+  testImplementation 'org.apache.commons:commons-lang3'
 
   errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
+++ b/keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
@@ -12,18 +12,15 @@
  */
 package tech.pegasys.signers.yubihsm2;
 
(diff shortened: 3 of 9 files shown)


keystorage/yubihsm2/build.gradle
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (C) 2020 ConsenSys AG.
+ *
+ * The source code is provided to licensees of PegaSys Plus for their convenience and internal
+ * business use. These files may not be copied, translated, and/or distributed without the express
+ * written permission of an authorized signatory for ConsenSys AG.
+ */
+
+apply plugin: 'java-library'
+apply plugin: 'java-test-fixtures'
```
rain-on, line 10: do you need test-fixtures?

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
```diff
@@ -0,0 +1,88 @@
...
+import org.assertj.core.api.Assumptions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+// manual dev test -
+// Make sure yubihsm2 is plugged in
+// for http/https url scheme, make sure connector daemon is running
+// for yhusb://serial=123456 url scheme, no connector daemon is required.
+// uncomment system properties in setup method and modify accordingly
+class YubiHsm2Test {
```
rain-on, line 28: is there a difference between YubiHsm, and YubiHsm2? (i.e. what is the 2 post-fix?)

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
```diff
@@ -0,0 +1,88 @@
...
+  private YubiHsm2 yubiHsm2;
+
+  @BeforeEach
+  void setup() {
+    // System.setProperty("yubihsmConnector", "yhusb://serial=123456");
```
rain-on, line 33: can the comments be removed? Or are they kept there as "samples/debug"?

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
```diff
@@ -0,0 +1,88 @@
...
+          "384a62688ee1d9a01c9d58e303f2b3c9bc1885e8131565386f75f7ae6ca8d147",
+          "4b6b5c682f2db7e510e0c00ed67ac896c21b847acadd8df29cf63a77470989d2",
+          "13086d684f4b1a1632178a8c5be08a2fb01287c4a78313c41373701eb8e66232",
+          "25296867ee96fa5b275af1b72f699efcb61586565d4c3c7e41f4b3e692471abd",
+          "10e1a313e573d96abe701d8848742cf88166dd2ded38ac22267a05d1d62baf71",
+          "0bdeebbad8f9b240192635c42f40f2d02ee524c5a3fe8cda53fb4897b08c66fe",
+          "5e8d5667ce78982a07242739ab03dc63c91e830c80a5b6adca777e3f216a405d"
+        };
+
+    final StopWatch stopWatch = new StopWatch();
```
rain-on, line 77: is the stop watch part of the test? If not - it should probably be removed (though I do like this stuff being atest - just not sure what to "assert")

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.TimeoutException;
+
+import org.apache.commons.io.IOUtils;
+import org.zeroturnaround.exec.ProcessExecutor;
+import org.zeroturnaround.exec.ProcessResult;
+
+/**
+ * Access secrets (BLS Keys) from Yubi HSM 2. https://developers.yubico.com/YubiHSM2/. It is assumed
```
jframe, line 28: Is this specific to BLS?

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
```diff
@@ -0,0 +1,88 @@
...
+            Optional.of(yubiShellPath),
+            caCert,
+            proxy);
+  }
+
+  @Test
+  void validDataIsReturnedFromYubiHSM() {
+    final String[] expectedKeys =
+        new String[] {
+          "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f",
```
jframe, line 64: are these keys are specific to your test device? is there a setup script to get these into the yubihsm?

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
jframe, line 28: This test should probably be marked as disabled so it doesn't get run as part of normal test run

gradle/versions.gradle
```diff
@@ -64,5 +68,9 @@ dependencyManagement {
     dependency 'org.web3j:core:4.5.14'
 
     dependency 'org.zeroturnaround:zt-exec:1.11'
+
+    dependency 'info.picocli:picocli:4.5.1'
+
+    dependency 'tech.pegasys.teku.internal:bls:0.12.6'
```
rain-on, line 74: don't think Signers can depend on Teku - this feels a bit backwards.

gradle/versions.gradle
```diff
@@ -64,5 +68,9 @@ dependencyManagement {
     dependency 'org.web3j:core:4.5.14'
 
     dependency 'org.zeroturnaround:zt-exec:1.11'
+
+    dependency 'info.picocli:picocli:4.5.1'
```
rain-on, line 72: don't think signers should depend upon Picocli - this should be a utility library

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+public class YubiHsm2 {
+  private final List<String> yubiHsmShellPathArgs;
+  private final String connectorUrl;
+  private final short authKey;
+  private final String password;
+  private final Optional<String> caCert;
+  private final Optional<String> proxy;
+  private final Optional<Map<String, String>> additionalEnvVars;
+
+  /**
```
rain-on, line 42: do we want to be doing javadoc for these functions, or does it just make our life harder in the long run? I.e. we barely do this anywhere - but maybe, given this is a library it makes sense

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+    this.authKey = authKeyObjId;
+    this.password = password;
+    this.caCert = caCert;
+    this.proxy = proxy;
+  }
+
+  /**
+   * Fetch opaque data from yubi hsm 2
+   *
+   * @param opaqueObjId The object id of opque
```
rain-on, line 73: "The Object Id of the opaque data to retrieve"?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+      return executeProcess(
+          getOpaqueDataArgs(opaqueObjId),
+          additionalEnvVars.orElse(Collections.emptyMap()),
+          password);
+    } catch (final IOException | TimeoutException | InterruptedException e) {
+      throw new YubiHsmException("Error in invoking yubihsm-shell process: " + e.getMessage());
+    }
+  }
+
+  static String executeProcess(
```
rain-on, line 88: is this used somewhere else? If not - there's no need to pass in the password and EnvVars, as they are part of this object (given that the cmdline args creation is not static, this probably doesn't need to be either).

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+    final ProcessResult processResult =
+        new ProcessExecutor()
+            .environment(additionalEnvVars)
+            .command(args)
+            .readOutput(true)
+            .redirectInput(IOUtils.toInputStream(password + "\n", StandardCharsets.UTF_8.name()))
+            .destroyOnExit()
+            .execute();
+
+    final List<String> linesAsUTF8 = processResult.getOutput().getLinesAsUTF8();
```
rain-on, line 100: linesAsUtf8? or what about utf8Lines? (just not capitalised UTF8)

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+   * @param caCert Optional CA Cert if connector is running in https mode
+   * @param proxy Optional proxy url if a proxy should be used to access connector url
+   */
+  public YubiHsm2(
+      final List<String> yubiHsmShellPathArgs,
+      final Optional<Map<String, String>> additionalEnvVars,
+      final String connectorUrl,
+      final short authKeyObjId,
+      final String password,
+      Optional<String> caCert,
```
rain-on, line 59: is this a caCert path, or the literal text of the caCert?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+      Optional<String> proxy) {
```
rain-on, line 60: could this be represented as a URL? I assume the purpose is that we can access a YubiHsm on a different computer?

rain-on: If this is specified, is the caCert required?

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
```diff
@@ -0,0 +1,178 @@
...
+            Optional.empty());
+
+    assertThatExceptionOfType(YubiHsmException.class)
+        .isThrownBy(() -> yubiHsm2.fetchKey((short) 11))
+        .withMessage("Unable to fetch data from YubiHSM: Failed to open Session");
+  }
+
+  private static void addOpaqueData() throws IOException, TimeoutException, InterruptedException {
+    for (int i = 1; i <= 10; i++) {
+      final BLSKeyPair blsKeyPair = BLSKeyPair.random(i);
```
rain-on, line 119: the storage doesn't need to be explicitly BLS oriented - it should be just random strings.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+    this.password = password;
+    this.caCert = caCert;
+    this.proxy = proxy;
+  }
+
+  /**
+   * Fetch opaque data from yubi hsm 2
+   *
+   * @param opaqueObjId The object id of opque
+   * @return The data stored against opaque object id (in hex format)
```
rain-on, line 74: are the secrets require to be strings? l.e _could_ the yubi store _anything_ - i.e what if its not a hex string, what if its my credit card number? 
I guess I mean - we should simple return the string value stored in the secret, unmodified (rather than stating it will be hex encoded).

gradle/versions.gradle
usmansaleem: it is used in tests.

gradle/versions.gradle
usmansaleem: it is also used in tests.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: I only do it for maintenance.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: 👍

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: Used in test ... I can potentially move it to a util class.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: auto-gen by IntelliJ. Let me fix it.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: Technically it is path, I used the same naming convention as yubishell was using. I'll modify it.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: no, both of them are optional.

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
usmansaleem: You are correct.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: yubi takes opaque data in following formats:
"base64", "binary", "PEM", "password", "hex", "ASCII". I think we can use ASCII instead of hex that I am currently using (as hex maps directly with BLS/SECP in hex encoding).

usmansaleem: We have to tell yubihsm which format we want the data to be stored and returned. The default for input and output is binary (i.e. reading from a file).

gradle/versions.gradle
usmansaleem: but since the data being stored doesn't need to be BLS, this can be removed.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
jframe, line 60: proxyUrl instead of proxy to make it clearer?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
```diff
@@ -0,0 +1,120 @@
...
+   * @throws YubiHsmException In case of errors in invoking process or if command returns error.
+   */
+  public String fetchKey(final short opaqueObjId) throws YubiHsmException {
```
jframe, line 77: Aren't we fetching the key value not the key itself?

gradle/versions.gradle
usmansaleem: Moved it directly into module's build file as it is mainly required at test runtime to execute the yubishell simulator.

gradle/versions.gradle
usmansaleem: removed.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: moved into a util class. Also added a comment that it is visible for testing. As guava is not on module's dependency, can't add the @VisibleForTest annotation.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: fixed as outputLines.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: renamed to `caCertPath`

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm2/YubiHsm2Test.java
usmansaleem: done.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: Refactored slightly so that we can pass optional output format. Defaults to `default`.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: agreed, refactored to `proxyUrl`.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm2/YubiHsm2.java
usmansaleem: you are right, renamed it to `fetchOpaqueData` which is exactly what it is doing.

keystorage/yubihsm2/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+  testImplementation 'org.assertj:assertj-core'
+  testImplementation 'org.junit.jupiter:junit-jupiter-engine'
+  testImplementation 'org.junit.jupiter:junit-jupiter-api'
+  testImplementation 'org.junit.jupiter:junit-jupiter-params'
+  testImplementation 'org.mockito:mockito-core'
+  testImplementation 'org.apache.commons:commons-lang3'
+  testImplementation 'com.google.guava:guava'
+
+  // required for resources/YubiShellSimulator.java
+  testRuntimeOnly 'info.picocli:picocli:4.5.1'
```
jframe, line 40: would prefer the version was in the versions gradle file

jframe (approved):

PR #88, opened 2020-10-23
YubiHsm using PKCS11

usmansaleem:
Added YubiHSM integration using PKCS11 module
Removed YubiHsm-Shell integration


keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
```diff
@@ -0,0 +1,140 @@
...
+  public Bytes fetchOpaqueData(final short opaqueObjId, final OpaqueDataFormat opaqueDataFormat)
+      throws YubiHsmException {
+    try {
+      initFind(opaqueObjId);
+
+      final byte[] data = findData(opaqueObjId);
+
+      return opaqueDataFormat == OpaqueDataFormat.HEX
+          ? Bytes.wrap(data)
+          : Bytes.fromHexString(new String(data, ISO_8859_1));
```
rain-on, line 58: hexstring must be a-f, 0-9 ... so should be representable using UTF_8? Or does this not really matter?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Override
+  public Bytes fetchOpaqueData(final short opaqueObjId, final OpaqueDataFormat opaqueDataFormat)
+      throws YubiHsmException {
+    try {
+      initFind(opaqueObjId);
+
+      final byte[] data = findData(opaqueObjId);
+
+      return opaqueDataFormat == OpaqueDataFormat.HEX
```
rain-on, line 56: think I'm missing something here - what does it mean if I'm asking for data in hex? Should 'HEX' be 'BINARY' instead?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
```diff
@@ -0,0 +1,140 @@
...
+      finalizeFind();
+    }
+  }
+
+  private byte[] findData(final short opaqueObjId) {
+    LOG.trace("Find Objects {}", opaqueObjId);
+
+    try {
+      final PKCS11Object[] data = session.findObjects(1);
+      if (data == null || data.length == 0) {
```
rain-on, line 70: can the data in the secret not be empty? I mean, its not useful, but it seems valid.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
```diff
@@ -0,0 +1,140 @@
...
+        Pkcs11ModuleFactory.initPkcs11Module(
+            configuration.getPkcs11ModulePath(), configuration.getPkcs11Conf());
+    this.session = Pkcs11SessionFactory.loginSession(module, configuration.getPin());
+  }
+
+  @Override
+  public Bytes fetchOpaqueData(final short opaqueObjId, final OpaqueDataFormat opaqueDataFormat)
+      throws YubiHsmException {
+    try {
+      initFind(opaqueObjId);
```
rain-on, line 52: the fact that initFind, and also findData take the same parameter, makes me wonder if both functions should appear on a separate object - like an OpaqueDataFinder?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
usmansaleem: it is not meant to be null (I think), but it can be 0 size (if opaque data doesn't exist).

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
usmansaleem: when we store opaque data in yubihsm, we can specify whether we are storing data as HEX (BINARY) or as ASCII (by setting `informat` in yubi-shell command, for instance). 

If we store a BLS key as Hex, it is stored directly as binary. If we store it as ASCII, Yubihsm stores the ascii value of the BLS key i.e. `5e` would be stored as `35 65`. 

Therefore, when we obtain back the data, it can either be stored as BLS binary ... or as Hex String.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
usmansaleem: it shouldn't matter.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Configuration.java
```diff
@@ -0,0 +1,55 @@
...
+    this.caCertPath = caCertPath;
+    this.proxyUrl = proxyUrl;
+  }
+
+  public Path getPkcs11ModulePath() {
+    return pkcs11ModulePath;
+  }
+
+  public char[] getPin() {
+    return String.format("%04X%s", authId, password).toCharArray();
```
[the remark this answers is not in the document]
usmansaleem: It is injected into Pkcs11 library (it takes a char[]).

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/YubiHsmSessionPkcs11.java
usmansaleem: we can possibly eliminate this step if we mandate that user must always use 'hex' input format to store BLS as Opaque data in YubiHsm for us to work with.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11ModuleFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.nio.file.Path;
+
+import iaik.pkcs.pkcs11.DefaultInitializeArgs;
+import iaik.pkcs.pkcs11.Module;
+import iaik.pkcs.pkcs11.TokenException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+class Pkcs11ModuleFactory {
```
jframe, line 26: Think class is misnamed, it's not creating a module. Perhaps Pkcs11ModuleProvider?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11SessionFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.logging.log4j.Logger;
+
+class Pkcs11SessionFactory {
+  private static final Logger LOG = LogManager.getLogger();
+
+  public static Session loginSession(final Module module, final char[] pin) {
+    final Session session = openReadOnlySession(getToken(module));
+    try {
+      session.login(Session.UserType.USER, pin);
+    } catch (final TokenException e) {
```
jframe, line 32: Also need to handle PKCS11Exception looks of the login method

keystorage/yubihsm2/src/test/java/tech/pegasys/signers/yubihsm/ManualYubiHsmPKCS11Test.java
```diff
@@ -0,0 +1,108 @@
...
+
+import java.nio.file.Path;
+import java.util.Optional;
+
+import org.apache.tuweni.bytes.Bytes;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+@Disabled("Require physical YubiHSM device")
+public class ManualYubiHsmPKCS11Test {
```
jframe, line 28: Is it possible to have a test using softhsm or something similar so have a test that always run?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/OpaqueDataFormat.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.signers.yubihsm;
+
+public enum OpaqueDataFormat {
+  ASCII,
+  HEX
```
rain-on, line 17: Wondering if we even want to differentiate this at this level. What is stored in the opaque data is Bytes (I assume) - if you as the client want to use it as ASCII, go for it - if its base64encoded, also fine .... i.e. the extraction maybe shouldn't be responsible for the conversion (that's application specific).

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/YubiHsm.java
```diff
@@ -0,0 +1,29 @@
...
+
+  /**
+   * Fetch key as opaque data from YubiHSM
+   *
+   * @param opaqueObjId Opaque object id
+   * @param opaqueDataFormat Specify format of stored data i.e. HEX or ASCII
+   * @return data as Bytes
+   * @throws YubiHsmException if unable to fetch data
+   */
+  Bytes fetchOpaqueData(short opaqueObjId, OpaqueDataFormat opaqueDataFormat)
```
rain-on, line 27: given the above comment, the format can/could be removed from here

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11Module.java
```diff
@@ -0,0 +1,76 @@
...
+   * @see <a href="https://developers.yubico.com/YubiHSM2/Component_Reference/PKCS_11/">YubiHSM
+   *     Configuration Options</a>
+   * @param pkcs11ModulePath The path to pkcs11 module .so or .dylib
+   * @param pkcs11InitConfig The pkcs11 module's initialization configuration string in lieu of
+   *     configuration file
+   */
+  public Pkcs11Module(final Path pkcs11ModulePath, final String pkcs11InitConfig) {
+    final Module module;
+    try {
+      module = Module.getInstance(pkcs11ModulePath.toString());
```
rain-on, line 42: can we put some debug/trace logging around here to indicate which module we're loading etc.? Such that if it fails, we can maybe work out why (not sure what data is attached to the IOException ... so this MAY be unrequired).

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11Module.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    LOG.debug("Initializing PKCS11 module with conf: {}", pkcs11InitConfig);
+    final DefaultInitializeArgs defaultInitializeArgs = new DefaultInitializeArgs();
+    defaultInitializeArgs.setReserved(pkcs11InitConfig);
+
+    try {
+      module.initialize(defaultInitializeArgs);
+    } catch (final TokenException e) {
+      LOG.error("Unable to initialize PKCS11 module {}", e.getMessage());
```
rain-on, line 54: if you're going to raise an exception with the same message, no need to log it as well.

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11Module.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * Create Pkcs11 Module.
+   *
+   * @see <a href="https://developers.yubico.com/YubiHSM2/Component_Reference/PKCS_11/">YubiHSM
+   *     Configuration Options</a>
+   * @param pkcs11ModulePath The path to pkcs11 module .so or .dylib
+   * @param pkcs11InitConfig The pkcs11 module's initialization configuration string in lieu of
+   *     configuration file
+   */
+  public Pkcs11Module(final Path pkcs11ModulePath, final String pkcs11InitConfig) {
```
rain-on, line 39: prefer this to be in a static creation function rather than a constructor, but not sure why - hangover from being old?

rain-on: For the record - some additional arguments for static creator function:
https://www.baeldung.com/java-constructors-vs-static-factory-methods

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11Module.java
```diff
@@ -0,0 +1,76 @@
...
+      module.initialize(defaultInitializeArgs);
+    } catch (final TokenException e) {
+      LOG.error("Unable to initialize PKCS11 module {}", e.getMessage());
+      throw new YubiHsmException("Unable to initialize PKCS11 module " + e.getMessage(), e);
+    }
+
+    this.module = module;
+  }
+
+  public Module getModule() {
```
rain-on, line 61: having a "getModule" basically opens this class up to be misused - i.e. people can now do bad things to your Module ... is there a better way we can abstact the Module? I.e. put the "openReadOnlySession" here, rather than exposing the module?

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11Session.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class Pkcs11Session implements AutoCloseable {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final Session session;
+
+  public Pkcs11Session(final Pkcs11Module module, final Pkcs11YubiHsmPin pin) {
+    final Session session = openReadOnlySession(getToken(module.getModule()));
```
rain-on, line 31: as before, constructors typically don't do "work" - they just set fields.
So either:
1. This needs to go into a static creator function, or
2. Pass in the session as a parameter to this, or
3. Both, pass the session as a parameter into a static creator

keystorage/yubihsm2/src/main/java/tech/pegasys/signers/yubihsm/pkcs11/Pkcs11Session.java
```diff
@@ -0,0 +1,101 @@
...
+      session.login(Session.UserType.USER, pin.getPin());
+    } catch (final TokenException e) {
+      LOG.error("YubiHSM Login failed {}", e.getMessage());
+      closeSession(session);
+      throw new YubiHsmException("Login Failed", e);
+    }
+    this.session = session;
+  }
+
+  public Session getSession() {
```
rain-on, line 43: Shouldn't just open up your session to everyone - removes benefits of encapsulation - can you move the users of this function into here?

jframe (approved):

PR #97, opened 2020-10-29
Keystore loading - make UUID and path optional

usmansaleem:
-- UUID and path are not used during decryption
 -- some other implementations don't generate them. This change make those keystores compatible.

### DIFF
--- a/bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/KeyStoreData.java
+++ b/bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/KeyStoreData.java
@@ -38,18 +38,18 @@ public class KeyStoreData {
   public KeyStoreData(
       @JsonProperty(value = "crypto", required = true) final Crypto crypto,
       @JsonProperty(value = "pubkey", required = true) final Bytes pubkey,
-      @JsonProperty(value = "path", required = true) final String path,
-      @JsonProperty(value = "uuid", required = true) final UUID uuid,
-      @JsonProperty(value = "version", required = true) final Integer version) {
+      @JsonProperty(value = "version", required = true) final Integer version,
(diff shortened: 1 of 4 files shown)


bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/KeyStoreData.java
```diff
@@ -38,18 +38,18 @@
   public KeyStoreData(
       @JsonProperty(value = "crypto", required = true) final Crypto crypto,
       @JsonProperty(value = "pubkey", required = true) final Bytes pubkey,
-      @JsonProperty(value = "path", required = true) final String path,
-      @JsonProperty(value = "uuid", required = true) final UUID uuid,
-      @JsonProperty(value = "version", required = true) final Integer version) {
+      @JsonProperty(value = "version", required = true) final Integer version,
+      @JsonProperty(value = "path", required = false) final String path,
```
rain-on, line 42: don't need to be explicit - required defaults to false.

bls-keystore/src/main/java/tech/pegasys/signers/bls/keystore/model/KeyStoreData.java
```diff
@@ -38,18 +38,18 @@
   public KeyStoreData(
```
rain-on, line 38: you may want a INCLUDE_NON_NULL - or whatever it is, such that when we generate a file, we don't put in null fields (i.e. the resulting json has the field missing, rather than field: null)

usmansaleem: for generation this constructor is not used. But you are right, let me add that annotation.

rain-on (approved):

PR #101, opened 2020-11-27
AzureKeyVault authentication using Managed Identity

usmansaleem:
Provide factory method to construct AzureKeyVault either using Client Secret credentials or Managed Identity Credentials.

### DIFF
--- a/keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
+++ b/keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
@@ -20,10 +20,11 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
+import com.azure.core.credential.TokenCredential;
 import com.azure.core.exception.ResourceNotFoundException;
 import com.azure.core.http.rest.PagedIterable;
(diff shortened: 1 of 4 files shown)


keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
```diff
@@ -83,12 +93,12 @@ public CryptographyClient fetchKey(final String keyName, final String keyVersion
...
 
     return new CryptographyClientBuilder()
-        .credential(clientSecretCredential)
+        .credential(tokenCredential)
         .keyIdentifier(keyId)
         .buildClient();
   }
 
-  public static String constructAzureKeyVaultUrl(final String keyVaultName) {
+  private static String constructAzureKeyVaultUrl(final String keyVaultName) {
```
rain-on, line 101: leaving this public makes it easier to use from other places - and given it doesn't expose anything "dangerous" internal to the class, better to make it open (IMHO).

rain-on (approved):

keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
usmansaleem: reverted.

PR #112, opened 2021-03-18
Update Azure Key Vault Signer signing algorithm

usmansaleem:
* Azure has deprecated curve name `SECP256K1`, instead they now use curve name `P-256K` when creating new keys or updating expired keys. 
* `ECDSA256` signature algorithm name has been deprecated. ES256K is to be used with `P-256K`.

### DIFF
--- a/signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/azure/AzureKeyVaultSignerFactory.java
+++ b/signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/azure/AzureKeyVaultSignerFactory.java
@@ -30,6 +30,7 @@ public class AzureKeyVaultSignerFactory {
   public static final String INACCESSIBLE_KEY_ERROR = "Failed to authenticate to vault.";
   public static final String INVALID_KEY_PARAMETERS_ERROR =
       "Keyvault does not contain key with specified parameters";
+  public static final String INVALID_CURVE_NAME = "Remote key has invalid curve name";
   private static final Logger LOG = LogManager.getLogger();
 
   private final boolean needsToHash;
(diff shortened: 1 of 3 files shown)


signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/azure/AzureKeyVaultSignerFactory.java
```diff
@@ -66,8 +67,14 @@ public Signer createSigner(final AzureConfig config) {
       throw new SignerInitializationException(INVALID_KEY_PARAMETERS_ERROR, e);
     }
     final JsonWebKey jsonWebKey = cryptoClient.getKey().getKey();
+    final String curveName = jsonWebKey.getCurveName().toString();
+    if (!curveName.equals("SECP256K1") && !curveName.equals("P-256K")) {
```
jframe, line 71: be nice to have constants for these strings

signing/secp256k1/impl/src/main/java/tech/pegasys/signers/secp256k1/azure/AzureKeyVaultSignerFactory.java
```diff
@@ -66,8 +67,14 @@ public Signer createSigner(final AzureConfig config) {
...
+      LOG.error("Unsupported curve name: {}. Expecting P-256K", curveName);
+      throw new SignerInitializationException(INVALID_CURVE_NAME);
+    }
     final Bytes rawPublicKey =
         Bytes.concatenate(Bytes.wrap(jsonWebKey.getX()), Bytes.wrap(jsonWebKey.getY()));
-    return new AzureKeyVaultSigner(config, rawPublicKey, needsToHash);
+    return new AzureKeyVaultSigner(
+        config, rawPublicKey, needsToHash, curveName.equals("SECP256K1"));
```
jframe, line 78: extracting curveName.equals("SECP256K1") as a variable say useDeprecated would make this clearer

jframe (approved):
LGTM

PR #115, opened 2021-09-16
Fix dependency issues

siladu:
jcenter/bintray is now defunct so removing repository references as it caused a build error.
Upgraded spotless and brought vertx versions inline with web3signer

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ import java.text.SimpleDateFormat
 
 plugins {
   id 'java-test-fixtures'
-  id 'com.diffplug.gradle.spotless' version '3.27.1'
+  id "com.diffplug.spotless" version "5.15.0"
   id 'com.github.ben-manes.versions' version '0.27.0'
   id 'com.github.hierynomus.license' version '0.15.0'
   id 'io.spring.dependency-management' version '1.0.9.RELEASE'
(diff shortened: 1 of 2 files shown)


usmansaleem (approved):
LGTM

build.gradle
```diff
@@ -17,7 +17,7 @@ import java.text.SimpleDateFormat
 
 plugins {
   id 'java-test-fixtures'
-  id 'com.diffplug.gradle.spotless' version '3.27.1'
+  id "com.diffplug.spotless" version "5.15.0"
```
jframe, line 20: nit: why the double quotes? other dependencies are using single quotes

gradle/versions.gradle
```diff
@@ -29,11 +29,13 @@ dependencyManagement {
 
     dependency 'io.rest-assured:rest-assured:4.2.0'
 
-    dependency 'io.vertx:vertx-codegen:3.8.2'
-    dependency 'io.vertx:vertx-core:3.8.2'
-    dependency 'io.vertx:vertx-unit:3.8.2'
-    dependency 'io.vertx:vertx-web-client:3.8.2'
-    dependency 'io.vertx:vertx-web:3.8.2'
+    dependencySet(group: 'io.vertx', version: '3.9.8') {
```
jframe, line 32: Looks like latest version is 3.9.9 https://vertx.io/blog/eclipse-vert-x-3-9-9/ can we update to that

build.gradle
siladu: ah, yep, copy paste :)

gradle/versions.gradle
siladu: @usmansaleem flagged this duration our conv yday to keep the version inline with web3signer... @usmansaleem is there specific reason for keeping them the same?

PR #117, opened 2021-09-22
Upgrade to Azure Key Vault 4.3.3 following vertx upgrade

siladu:
**NOTE: Azure tests currently failing due to potential issue with config in key vault**

Remove support for previously deprecated SECP256K1 curve.
'TestKey' updated in vault to use P-256K instead of SECP256K1.
Blank key-version defaults to the latest key-version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 #Changelog
 
+## 1.0.18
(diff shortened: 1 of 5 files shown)


siladu:
Replaced by https://github.com/ConsenSys/signers/pull/118

PR #118, opened 2021-09-23
Upgrade to Azure Key Vault 4.3.3 following vertx upgrade

siladu:
Remove support for previously deprecated SECP256K1 curve.
Use 'TestKey2' in Azure Key Vault to use P-256K instead of SECP256K1.
Blank key-version defaults to the latest key-version.


jframe (approved):
LGTM

PR #119, opened 2021-10-26
load signers based on signer identifier in filename

usmansaleem:


acceptance-tests/src/testFixtures/java/tech.pegasys.signers.secp256k1/DefaultFileSelector.java
```diff
@@ -33,4 +33,19 @@
...
+  @Override
+  public Filter<Path> getSingleConfigFileFilter(final String address) {
+    // remove 0x if part of address
+    return entry -> {
+      final String fileName;
+      if (address.toUpperCase().startsWith("0X")) {
+        fileName = address.substring(2);
+      } else {
+        fileName = address;
+      }
```
jframe, line 46: This looks like an improved version of what the getSpecificConfigFileFilter can we use the same code for both i.e. removing the 0x prefix if needed.

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SignerProvider.java
```diff
@@ -20,6 +20,11 @@
 
   Optional<Signer> getSigner(ECPublicKey publicKey);
 
+  default Optional<Signer> getSigner(String address) {
```
jframe, line 23: Is address always an ethereum address? How will this work for filecoin addresses? Is this just the string representation of a filecoin address?

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SignerProvider.java
```diff
@@ -20,6 +20,11 @@
...
+    // not applicable for Single Signer Provider mode
+    throw new UnsupportedOperationException("getSigner by address not supported");
```
jframe, line 25: Think this should also work for the single signer provider, it's just an alternative to using the public key

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SignerProvider.java
```diff
@@ -15,12 +15,13 @@
...
 import java.util.Set;
+import java.util.function.Function;
 
 public interface SignerProvider {
 
-  Optional<Signer> getSigner(ECPublicKey publicKey);
+  Optional<Signer> getSigner(SignerIdentifier publicKey);
 
-  Set<ECPublicKey> availablePublicKeys();
+  Set<ECPublicKey> availablePublicKeys(Function<ECPublicKey, SignerIdentifier> identifierFunction);
```
jframe, line 24: This makes the API a bit odd and this function doesn't seem related to returning available public keys. I can see why it's necessary though from how it's used in the MultiKeySignerProvider.

jframe (approved):

signing/secp256k1/api/src/main/java/tech/pegasys/signers/secp256k1/api/SignerProvider.java
usmansaleem: Yes, I couldn't think of another way to validate that the loaded signers matches the "identifier" in filename without this approach.

PR #123, opened 2021-11-26
Allow key store data to be loaded from string.

rolfyone:


bls-keystore/src/test/java/tech/pegasys/signers/bls/keystore/KeyStoreTest.java
```diff
@@ -113,6 +114,11 @@ void loadSCryptKeyStoreAndTestInvalidPassword() {
     invalidPasswordValidation(SCRYPT_KEYSTORE_RESOURCE);
   }
 
+  @Test
```
jframe, line 117: Making these parameterised tests would probably this a bit cleaner

jframe (approved):
LGTM

PR #129, opened 2022-01-17
Update Vertx to 4.x and various dependencies 

usmansaleem:
Update Vertx to 4.x and various other dependencies to latest versions. Make many signers modules backward incompatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## 2.0.0
+### Features Added
+- Update various dependencies. 
+- Adding Vertx 4 which makes signers backward incompatible. 
+
+---
 ## 1.0.25
 ### Features Added
 - Update Tuweni libraries to 2.0.0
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
 # Changelog
+## 2.0.0
```
jframe, line 2: Does it make sense to bump this to a major version?

gradle/license-report-config/allowed-licenses.json
```diff
@@ -63,6 +63,9 @@
     {
       "moduleLicense": "IAIK of Graz University of Technology License",
       "moduleName": "org.xipki.iaik:sunpkcs11-wrapper"
+    },
+    {
```
jframe, line 67: Is this needed? It doesn't seem to define an allowed license

usmansaleem: yes, a peculiarity of the license plugin we are using. Since it can't determine the type of license, we add the license name in allowed list. Then we add it in override list to use our supplied license name.

CHANGELOG.md
usmansaleem: yes, due to vertx 4, a lot of modules become backward-incompatible. Hence it is a good idea to bump the major version as an indication for such.

jframe (approved):
LGTM

PR #134, opened 2022-02-02
Fix Azure AT test

jframe:
Fix Azure AT test. The error message seems to somewhat inconsistent so have made the test less brittle.


siladu (approved):
LGTM

PR #135, opened 2022-02-07
AWS keystore retrieval with secrets manager

georgep9:
## PR description

Class for retrieving BLS keystores from AWS Secrets manager.

## Fixed Issue(s)
Fixes https://github.com/ConsenSys/signers/issues/131


keystorage/aws/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+
+testFixturesJar {
+  archiveBaseName = calculateJarName(project)
+}
+
+dependencies {
+
+  implementation platform('software.amazon.awssdk:bom:2.17.125')
+
+  implementation 'io.vertx:vertx-core'
```
jframe, line 37: Looks like this only needed for parsing the json object? If that's the case it would be better to just use Jackson which Vertx is using under the hood anyway.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public static AwsSecretsManager createAwsSecretsManager(
+      final String accessKeyId, final String secretAccessKey, final String region) {
+    final AwsBasicCredentials awsBasicCredentials =
+        AwsBasicCredentials.create(accessKeyId, secretAccessKey);
+    final StaticCredentialsProvider credentialsProvider =
+        StaticCredentialsProvider.create(awsBasicCredentials);
+
+    final SecretsManagerClient secretsManagerClient =
```
jframe, line 41: What's the lifecycle of the SecretsManagerClient do we need to call close on this at some point?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public Optional<String> fetchSecret(final String secretName) {
+    try {
+      final GetSecretValueRequest getSecretValueRequest =
+          GetSecretValueRequest.builder().secretId(secretName).build();
+      final GetSecretValueResponse valueResponse =
+          secretsManagerClient.getSecretValue(getSecretValueRequest);
+      return Optional.of(valueResponse.secretString());
+    } catch (SecretsManagerException e) {
```
jframe, line 64: If key doesn't exist what exception is thrown? Is it the ResourceNotFoundException? It might be handled somewhere but expected to see Optional.empty be returned if some not found exception was thrown.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public static AwsSecretsManager createAwsSecretsManager(final String region) {
+    final SecretsManagerClient secretsManagerClient =
+        SecretsManagerClient.builder().region(Region.of(region)).build();
+
+    return new AwsSecretsManager(secretsManagerClient);
+  }
+
+  public Optional<String> fetchSecret(final String secretName) {
```
jframe, line 57: This string is actually json? Seems like this should be private internal method.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private AwsSecretsManager awsSecretsManagerDefault;
+  private AwsSecretsManager awsSecretsManagerExplicit;
+
+  @BeforeAll
+  void setupSecretManagers() {
+    final String awsRegion = "us-east-2";
+    final String awsAccessKeyId = System.getenv("AWS_ACCESS_KEY_ID");
+    final String awsSecretAccessKey = System.getenv("AWS_SECRET_ACCESS_KEY");
+    this.awsSecretsManagerDefault = AwsSecretsManager.createAwsSecretsManager(awsRegion);
```
jframe, line 41: Be good if to do something similar to the Azure test probably using Junit Assumptions where by if these environment variables don't exist we ignore the test.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Test
+  void fetchingNonExistentSecretThrowsException() {
+    assertThatExceptionOfType(RuntimeException.class)
+        .isThrownBy(() -> awsSecretsManagerDefault.fetchSecret("empty"))
+        .withMessageContaining("Secrets Manager can't find the specified secret.");
+  }
+
+  @Test
+  void fetchingNonExistentSecretValueReturnsNull() {
```
jframe, line 66: nit: This is actually testing that secret is empty not null

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Test
+  void fetchSecretValueWithExplicitManager() {
+    Optional<String> secret = awsSecretsManagerExplicit.fetchSecretValue(SECRET_NAME, SECRET_KEY);
+    assertThat(secret).hasValue(EXPECTED_KEYSTORE);
+  }
+
+  @Test
+  void fetchingNonExistentSecretThrowsException() {
+    assertThatExceptionOfType(RuntimeException.class)
```
jframe, line 60: In this case I we should return an empty secret not an error as this what we did with the Azure API

keystorage/azure/src/test/java/tech/pegasys/signers/azure/AzureKeyVaultTest.java
```diff
@@ -55,7 +55,7 @@ void connectingWithInvalidClientSecretThrowsException() {
         createUsingClientSecretCredentials(CLIENT_ID, "invalid", TENANT_ID, VAULT_NAME);
     assertThatExceptionOfType(RuntimeException.class)
         .isThrownBy(() -> azureKeyVault.fetchSecret(SECRET_NAME))
-        .withMessageContaining("Invalid client secret");
+        .withMessageContaining("Invalid client secret is provided");
```
jframe, line 58: Can we not change this back. Deliberately shorted this as the full message we was inconsistent sometimes it was returning "Invalid client secret is provided" sometimes "Invalid client secret provided". But that "Invalid client secret" prefix was consistent.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
lucassaldanha: It implements the `SkAutoCloseable` interface that has a `close()` method so we might need to look at how this should be handled.

https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/secretsmanager/SecretsManagerClient.html

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -0,0 +1,73 @@
...
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest;
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
+import software.amazon.awssdk.services.secretsmanager.model.SecretsManagerException;
+
+public class AwsSecretsManager implements Closeable {
+
+  private final SecretsManagerClient secretsManagerClient;
+
+  private AwsSecretsManager(SecretsManagerClient secretsManagerClient) {
```
jframe, line 30: nit: final

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public Optional<String> fetchSecret(final String secretName) {
+    try {
+      final GetSecretValueRequest getSecretValueRequest =
+          GetSecretValueRequest.builder().secretId(secretName).build();
+      final GetSecretValueResponse valueResponse =
+          secretsManagerClient.getSecretValue(getSecretValueRequest);
+      return Optional.of(valueResponse.secretString());
+    } catch (SecretsManagerException e) {
+      throw new RuntimeException(e.awsErrorDetails().errorMessage());
```
jframe, line 65: Can we include an error message here as well along with the exception.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -0,0 +1,132 @@
...
+    Optional<String> secret = awsSecretsManagerDefault.fetchSecret(secretName);
+    assertThat(secret).hasValue(SECRET_VALUE);
+  }
+
+  @Test
+  void fetchSecretWithExplicitManager() {
+    Optional<String> secret = awsSecretsManagerExplicit.fetchSecret(secretName);
+    assertThat(secret).hasValue(SECRET_VALUE);
+  }
+
```
jframe, line 125: Can we test what happens if we connection details?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public Optional<String> fetchSecret(final String secretName) {
+    try {
+      final GetSecretValueRequest getSecretValueRequest =
+          GetSecretValueRequest.builder().secretId(secretName).build();
+      final GetSecretValueResponse valueResponse =
+          secretsManagerClient.getSecretValue(getSecretValueRequest);
+      return Optional.of(valueResponse.secretString());
+    } catch (RuntimeException e) {
```
jframe, line 63: We don't want to treat all runtime errors as not found just the ResourceNotFoundException, the other exception we to continue allowing to bubble up.

jframe (approved):

Issue #138, opened 2022-03-11
Remove explicit AWS region when initializing default environment authenticated AwsSecretsManager


georgep9:
The AWS SDK should be able to authenticate a secrets manager connection using the [default region provider chain](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html)

**Tasks**
* Implement createAwsSecretsManager method (no params) that initializes and returns a SecretsManager connection using the environment's region.

PR #139, opened 2022-03-16
Cacheable AWS Secrets Manager

georgep9:
Provider class to cache AWS secrets managers based on access key ID. 

Fixes issue #140.
Additional changes are included to fix issue #138 (removed explicit Region for default manager).

### DIFF
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -48,9 +48,8 @@ public class AwsSecretsManager implements Closeable {
     return new AwsSecretsManager(secretsManagerClient);
   }
 
-  public static AwsSecretsManager createAwsSecretsManager(final String region) {
-    final SecretsManagerClient secretsManagerClient =
-        SecretsManagerClient.builder().region(Region.of(region)).build();
+  public static AwsSecretsManager createAwsSecretsManager() {
+    final SecretsManagerClient secretsManagerClient = SecretsManagerClient.builder().build();
 
     return new AwsSecretsManager(secretsManagerClient);
   }
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -12,9 +12,6 @@
  */
 package tech.pegasys.signers.aws;
 
-import java.io.Closeable;
-import java.util.Optional;
-
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
@@ -24,7 +21,10 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRespon
 import software.amazon.awssdk.services.secretsmanager.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.secretsmanager.model.SecretsManagerException;
 
-public class AwsSecretsManager implements Closeable {
+import java.io.Closeable;
+import java.util.Optional;
+
+class AwsSecretsManager implements Closeable {
 
   private final SecretsManagerClient secretsManagerClient;
 
@@ -32,7 +32,7 @@ public class AwsSecretsManager implements Closeable {
     this.secretsManagerClient = secretsManagerClient;
   }
 
-  public static AwsSecretsManager createAwsSecretsManager(
+  static AwsSecretsManager createAwsSecretsManager(
       final String accessKeyId, final String secretAccessKey, final String region) {
     final AwsBasicCredentials awsBasicCredentials =
         AwsBasicCredentials.create(accessKeyId, secretAccessKey);
@@ -48,7 +48,7 @@ public class AwsSecretsManager implements Closeable {
     return new AwsSecretsManager(secretsManagerClient);
   }
 
-  public static AwsSecretsManager createAwsSecretsManager() {
+  static AwsSecretsManager createAwsSecretsManager() {
     final SecretsManagerClient secretsManagerClient = SecretsManagerClient.builder().build();
 
     return new AwsSecretsManager(secretsManagerClient);
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -24,7 +24,7 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRespon
 import software.amazon.awssdk.services.secretsmanager.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.secretsmanager.model.SecretsManagerException;
 
-class AwsSecretsManager implements Closeable {
+public class AwsSecretsManager implements Closeable {
 
   private final SecretsManagerClient secretsManagerClient;
 
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -12,6 +12,9 @@
  */
 package tech.pegasys.signers.aws;
 
+import java.io.Closeable;
+import java.util.Optional;
+
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
@@ -20,9 +23,6 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueReques
 import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
 import software.amazon.awssdk.services.secretsmanager.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.secretsmanager.model.SecretsManagerException;
-
-import java.io.Closeable;
-import java.util.Optional;
 
 class AwsSecretsManager implements Closeable {
 
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManager.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2022 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.signers.aws;
+
+import java.util.HashMap;
+
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+
+public class CacheableAwsSecretsManager {
+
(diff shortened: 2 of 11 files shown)


keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManager.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ExecutionException;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+
+public class CacheableAwsSecretsManager {
+
+  private static final Cache<String, AwsSecretsManager> awsSecretsManagerCache =
+      CacheBuilder.newBuilder().build();
```
jframe, line 25: Should set a maximum size for the cache, we don't want this unbounded

keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManager.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+
+public class CacheableAwsSecretsManager {
+
+  private static final Cache<String, AwsSecretsManager> awsSecretsManagerCache =
```
jframe, line 24: Still don't like the static field here. Think it's safe because of the concurrent hashmap but we will still need be careful how use the CacheableAwsSecretsManager. E.g. what happens if a reload is happening at the same time as the initial key loading do we need to be careful when we call clearCache?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class CacheableAwsSecretsManagerTest {
+
+  private final String RW_AWS_ACCESS_KEY_ID = System.getenv("RW_AWS_ACCESS_KEY_ID");
```
siladu, line 26: What does RW stand for?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Test
+  void validateClearCache() {
+    final AwsSecretsManager awsSecretsManager = createSpecifiedSecretsManager();
+    final AwsSecretsManager awsSecretsManagerRW = createSpecifiedSecretsManagerRW();
+    CacheableAwsSecretsManager.clearCache();
+    assertThat(createSpecifiedSecretsManager()).isNotEqualTo(awsSecretsManager);
+    assertThat(createSpecifiedSecretsManagerRW()).isNotEqualTo(awsSecretsManagerRW);
+  }
+}
```
siladu, line 96: Do we need a test to verify that two different regions don't use the same object as well?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @AfterEach
+  void teardown() {
+    CacheableAwsSecretsManager.clearCache();
+  }
+
+  @Test
+  void returnsCachedSpecifiedSecretsManager() {
+    assertThat(createSpecifiedSecretsManager()).isEqualTo(createSpecifiedSecretsManager());
```
siladu, line 68: This turned out neat :)

I think a slight improvement is to use `isSameAs` since that will always check object equality (reference) instead of relying on the overrideable AwsSecretsManager.equals method.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
georgep9: Read/Write

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
georgep9: Not sure. I've left it out because the current test pipeline supports services within the us-east-2 region. 

From a security standpoint, should a web3signer client be deployed and used to fetch secrets from multiple regions?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
georgep9: That would be even neater :)

keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProvider.java
```diff
@@ -0,0 +1,58 @@
...
+        accessKeyId,
+        () -> AwsSecretsManager.createAwsSecretsManager(accessKeyId, secretAccessKey, region));
+  }
+
+  public AwsSecretsManager createAwsSecretsManager() {
+    final String key = DefaultCredentialsProvider.create().resolveCredentials().accessKeyId();
+    return fromCacheOrCallable(key, () -> AwsSecretsManager.createAwsSecretsManager());
+  }
+
+  public void clearCache() {
```
usmansaleem, line 51: 1. Consider implement Closeable interface
2. Try to consume runtime exception from aws close method so that its a clean close. (logging it though)

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProviderTest.java
```diff
@@ -0,0 +1,104 @@
...
+    initializeCacheableAwsSecretsManagerProvider();
+  }
+
+  @AfterEach
+  void teardown() {
+    cacheableAwsSecretsManagerProvider.close();
+  }
+
+  @Test
+  void isSameAsCachedSpecifiedSecretsManager() {
```
jframe, line 75: Would be good to have a variant of this, that with multiple items in the cache it returns the correct cached secrets manager

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerTest.java
siladu: What's the significance of read/write to these tests?

To me, it looks like we just want a second instance of the config...would e.g. `ANOTHER_AWS_ACCESS_KEY_ID` suffice?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProvider.java
```diff
@@ -39,20 +43,31 @@ private AwsSecretsManager fromCacheOrCallable(
...
+        new AwsKeyIdentifier(accessKeyId, region),
         () -> AwsSecretsManager.createAwsSecretsManager(accessKeyId, secretAccessKey, region));
   }
 
   public AwsSecretsManager createAwsSecretsManager() {
-    final String key = DefaultCredentialsProvider.create().resolveCredentials().accessKeyId();
-    return fromCacheOrCallable(key, () -> AwsSecretsManager.createAwsSecretsManager());
+    final String accessKeyId =
+        DefaultCredentialsProvider.create().resolveCredentials().accessKeyId();
+    final String region = DefaultAwsRegionProviderChain.builder().build().getRegion().toString();
```
siladu, line 53: What's the reasoning for using the String instead of passing in Region?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProvider.java
```diff
@@ -17,20 +17,24 @@
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
 
 public class CacheableAwsSecretsManagerProvider {
```
siladu, line 25: Wondering now is the time to merge CacheableAwsSecretsManagerProvider into AwsSecretsManager as we previously discussed?

I don't think there's a reason to have the non-cacheable version (if we _do_ want to provide a choice of two, then I think the API should be normalised rather than having two patterns for constructing AwsSecretsManager).

keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProvider.java
```diff
@@ -0,0 +1,76 @@
...
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.regions.providers.DefaultAwsRegionProviderChain;
+
+public class CacheableAwsSecretsManagerProvider implements Closeable {
```
jframe, line 27: Suggest rename this to AwsSecretsManagerProvider

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -48,9 +48,8 @@ public static AwsSecretsManager createAwsSecretsManager(
     return new AwsSecretsManager(secretsManagerClient);
   }
 
-  public static AwsSecretsManager createAwsSecretsManager(final String region) {
-    final SecretsManagerClient secretsManagerClient =
-        SecretsManagerClient.builder().region(Region.of(region)).build();
+  public static AwsSecretsManager createAwsSecretsManager() {
```
jframe, line 51: Can you reduce the scope of this and the other create method to package level so that only CacheableAwsSecretsManagerProvider can create the AwsSecretsManager

keystorage/aws/src/main/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProvider.java
jframe: Agree, there is no need to have a non-cacheable version.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/CacheableAwsSecretsManagerProviderTest.java
```diff
@@ -0,0 +1,125 @@
...
+    assertThat(createSpecifiedSecretsManager()).isSameAs(createSpecifiedSecretsManager());
+  }
+
+  @Test
+  void isSameAsCachedDefaultSecretsManager() {
+    assertThat(createDefaultSecretsManager()).isSameAs(createDefaultSecretsManager());
+  }
```
siladu, line 95: How about a test like this?

```suggestion
  }
  
  @Test
  void secretsManagerIsNotCachedWhenCacheSizeIsSetToZero() {
    cacheableAwsSecretsManagerProvider = new CacheableAwsSecretsManagerProvider(0);
    assertThat(createSpecifiedSecretsManager()).isNotSameAs(createSpecifiedSecretsManager());
  }
```

Note, you'd have to move `initializeCacheableAwsSecretsManagerProvider()` instead a @BeforeEach instead of @BeforeAll for this to work

siladu (approved):
LGTM

PR #141, opened 2022-03-26
Support for AWS bulk loading

georgep9:
Implemented mapper function similar to Azure to support bulk loading of secrets from AWS Secrets Manager.

Supports issue https://github.com/ConsenSys/web3signer/issues/499

### DIFF
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -32,8 +32,8 @@ import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueReques
 import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
 import software.amazon.awssdk.services.secretsmanager.model.ListSecretsRequest;
 import software.amazon.awssdk.services.secretsmanager.model.ResourceNotFoundException;
-import software.amazon.awssdk.services.secretsmanager.model.SecretListEntry;
 import software.amazon.awssdk.services.secretsmanager.model.SecretsManagerException;
+import software.amazon.awssdk.services.secretsmanager.paginators.ListSecretsIterable;
 
 public class AwsSecretsManager implements Closeable {
 
@@ -81,7 +81,7 @@ public class AwsSecretsManager implements Closeable {
     }
   }
 
-  private Collection<SecretListEntry> listSecrets(
+  private ListSecretsIterable listSecrets(
       final List<String> tagKeys, final List<String> tagValues) {
     final ListSecretsRequest.Builder listSecretsRequestBuilder = ListSecretsRequest.builder();
     if (!tagKeys.isEmpty()) {
@@ -92,8 +92,7 @@ public class AwsSecretsManager implements Closeable {
       listSecretsRequestBuilder.filters(
           Filter.builder().key(FilterNameStringType.TAG_VALUE).values(tagValues).build());
     }
-    final ListSecretsRequest listSecretsRequest = listSecretsRequestBuilder.build();
-    return secretsManagerClient.listSecrets(listSecretsRequest).secretList();
+    return secretsManagerClient.listSecretsPaginator(listSecretsRequestBuilder.build());
   }
 
   public <R> Collection<R> mapSecrets(
@@ -102,20 +101,30 @@ public class AwsSecretsManager implements Closeable {
       final BiFunction<String, String, R> mapper) {
     final Set<R> result = ConcurrentHashMap.newKeySet();
     listSecrets(tagKeys, tagValues)
-        .parallelStream()
-        .forEach(
-            secretEntry -> {
-              try {
-                final String secretValue = fetchSecret(secretEntry.name()).get();
-                final R obj = mapper.apply(secretEntry.name(), secretValue);
-                if (obj != null) {
-                  result.add(obj);
-                } else {
-                  LOG.warn("Mapped '{}' to a null object, and was discarded", secretEntry.name());
-                }
-              } catch (final Exception e) {
-                LOG.warn("Failed to map secret '{}' to requested object type.", secretEntry.name());
-              }
+        .iterator()
+        .forEachRemaining(
+            listSecretsResponse -> {
+              listSecretsResponse
+                  .secretList()
+                  .parallelStream()
+                  .forEach(
+                      secretEntry -> {
+                        try {
+                          final String secretValue = fetchSecret(secretEntry.name()).get();
+                          final R obj = mapper.apply(secretEntry.name(), secretValue);
+                          if (obj != null) {
+                            result.add(obj);
+                          } else {
+                            LOG.warn(
+                                "Mapped '{}' to a null object, and was discarded",
+                                secretEntry.name());
+                          }
+                        } catch (final Exception e) {
+                          LOG.warn(
+                              "Failed to map secret '{}' to requested object type.",
+                              secretEntry.name());
+                        }
+                      });
             });
     return result;
   }
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -14,7 +14,6 @@ package tech.pegasys.signers.aws;
 
 import java.io.Closeable;
 import java.util.Collection;
-import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -82,7 +81,7 @@ public class AwsSecretsManager implements Closeable {
   }
 
   private ListSecretsIterable listSecrets(
-      final List<String> tagKeys, final List<String> tagValues) {
+      final Collection<String> tagKeys, final Collection<String> tagValues) {
     final ListSecretsRequest.Builder listSecretsRequestBuilder = ListSecretsRequest.builder();
     if (!tagKeys.isEmpty()) {
       listSecretsRequestBuilder.filters(
@@ -96,8 +95,8 @@ public class AwsSecretsManager implements Closeable {
   }
 
   public <R> Collection<R> mapSecrets(
-      final List<String> tagKeys,
-      final List<String> tagValues,
+      final Collection<String> tagKeys,
+      final Collection<String> tagValues,
       final BiFunction<String, String, R> mapper) {
     final Set<R> result = ConcurrentHashMap.newKeySet();
     listSecrets(tagKeys, tagValues)
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -79,6 +79,12 @@ public class AwsSecretsManager implements Closeable {
     }
   }
 
+  public List<SecretListEntry> getAvailableSecrets(final String secretNamePrefix) {
+    return getAvailableSecrets().stream()
+        .filter(s -> !s.arn().startsWith(secretNamePrefix))
+        .collect(Collectors.toList());
+  }
+
   @Override
   public void close() {
     this.secretsManagerClient.close();
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
@@ -12,14 +12,6 @@
  */
 package tech.pegasys.signers.aws;
 
-import java.io.Closeable;
-import java.util.Collection;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.BiFunction;
-
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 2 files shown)


keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +76,39 @@ static AwsSecretsManager createAwsSecretsManager() {
     }
   }
 
+  public <R> Collection<R> mapSecretsList(
```
jframe, line 79: Just mapSecrets then it would be consistent with what we are using in AzureKeyVault

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +76,39 @@ static AwsSecretsManager createAwsSecretsManager() {
...
+      final String prefix, final BiFunction<String, String, R> mapper) {
+    final Set<R> result = ConcurrentHashMap.newKeySet();
+    secretsManagerClient.listSecretsPaginator().stream()
```
jframe, line 82: What is default paging size? If the default is less than the max 100 we should configure a custom ListSecretsRequest to minimised the number additional requests needed

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +76,39 @@ static AwsSecretsManager createAwsSecretsManager() {
...
+        .parallel()
+        .forEach(
+            p ->
+                p.secretList()
+                    .parallelStream()
+                    .map(secretEntry -> secretEntry.name())
+                    .filter(secretName -> secretName.startsWith(prefix))
```
jframe, line 89: This seems like an odd filter to have. What would make sense from how AWS is used? I think that tag might be more appropriate. But also supporting having no filter at all.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -143,4 +158,81 @@ void fetchingNonExistentSecretReturnsEmpty() {
...
+    final Optional<AbstractMap.SimpleEntry<String, String>> secretEntry =
+        secretEntries.stream().filter(e -> e.getKey().equals(secretName)).findAny();
+    assertThat(secretEntry).isPresent();
+    assertThat(secretEntry.get().getValue()).isEqualTo(SECRET_VALUE);
+  }
+
+  @Test
+  void multipleSecretsCanBeFetchedAndMapped() {
+    createSecret(); // 2nd
+    createSecret(); // 3rd
```
jframe, line 174: Why is this the 3rd secret? I only see two secrets created in this test

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -143,4 +158,81 @@ void fetchingNonExistentSecretReturnsEmpty() {
...
+    final Collection<AbstractMap.SimpleEntry<String, String>> secretEntries =
+        awsSecretsManagerExplicit.mapSecretsList(AbstractMap.SimpleEntry::new);
+
+    secretNames.forEach(secretName -> validateMappedSecret(secretEntries, secretName));
+  }
+
+  @Test
+  void multipleSecretsWithoutPrefixCanBeFetchedAndMapped() {
+    secretNamePrefix = "";
+    createSecret(); //
```
jframe, line 185: Empty comment can be removed

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -143,4 +158,81 @@ void fetchingNonExistentSecretReturnsEmpty() {
...
+
+  @Test
+  void multipleSecretsCanBeFetchedAndMapped() {
+    createSecret(); // 2nd
+    createSecret(); // 3rd
+
+    final Collection<AbstractMap.SimpleEntry<String, String>> secretEntries =
+        awsSecretsManagerExplicit.mapSecretsList(AbstractMap.SimpleEntry::new);
+
+    secretNames.forEach(secretName -> validateMappedSecret(secretEntries, secretName));
```
jframe, line 179: Can the size also be checked? I'm a bit dubious that this does this actually multiple secrets as the same key is backed by a set in the mapping function.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -143,4 +158,81 @@ void fetchingNonExistentSecretReturnsEmpty() {
...
+  void multipleSecretsWithoutPrefixCanBeFetchedAndMapped() {
+    secretNamePrefix = "";
+    createSecret(); //
+    createSecret(); //
+
+    final Collection<AbstractMap.SimpleEntry<String, String>> secretEntries =
+        awsSecretsManagerExplicit.mapSecretsList(AbstractMap.SimpleEntry::new);
+
+    secretNames.forEach(secretName -> validateMappedSecret(secretEntries, secretName));
+    secretNamePrefix = "signers-aws-integration/";
```
jframe, line 192: why does this need assigning at the end of the test?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -45,6 +51,8 @@
   private AwsSecretsManager awsSecretsManagerInvalidCredentials;
   private SecretsManagerClient secretsManagerClient;
   private String secretName;
+  private List<String> secretNames = new ArrayList<>();
```
jframe, line 54: this can be final

jframe, line 54: is this field needed? could a local variable be used in the tests instead

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
georgep9: Might be tricky with other unit tests running in parallel. Although more unique secret names per unit test would resolve that.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
georgep9: reassign `secretNamePrefix` back to its default value `"signers-aws-integration/"`

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
georgep9: The current signers read-only IAM policy used for the unit tests have restricted permissions and can only read secrets that have a resource/ARN under `arn:aws:secretsmanager:*:*:secret:signers-aws-integration/*`.
This appears to be the common approach for grouping secrets for specific use cases, hence the filter implemented in this method.

georgep9: I agree that tags would be more appropriate

georgep9:
perhaps requires additional unit tests;
```java
listSecretsAndFetchSecretsWithSingleSecretSingleTag
listSecretsAndFetchSecretsWithSingleSecretMultipleTags
listSecretsAndFetchSecretsWithMultipleSecretsSingleTag
listSecretsAndFetchSecretsWithMultipleSecretsMultipleTags
```

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +80,39 @@ public Optional<String> fetchSecret(final String secretName) {
     }
   }
 
+  private Collection<SecretListEntry> listSecrets(final AbstractMap<String, String> tags) {
+    final ListSecretsRequest.Builder listSecretsRequestBuilder = ListSecretsRequest.builder();
+    if (tags != null) {
+      final Filter keys = Filter.builder().key("tag-key").values(tags.keySet()).build();
+      final Filter values = Filter.builder().key("tag-value").values(tags.values()).build();
```
jframe, line 87: What do we do if we only want to filter by tag key and not tag value? How is this represented? Would be better to just have two lists of tag keys and tag values we want to filter on?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +80,39 @@ public Optional<String> fetchSecret(final String secretName) {
...
+      listSecretsRequestBuilder.filters(keys, values);
+    }
+    final ListSecretsRequest listSecretsRequest = listSecretsRequestBuilder.build();
```
jframe, line 90: Should we also set the maxResults as well? Not sure what the default is but don't think it is the maximum of 100.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +80,39 @@ public Optional<String> fetchSecret(final String secretName) {
...
+    return secretsManagerClient.listSecrets(listSecretsRequest).secretList();
```
jframe, line 91: How does this compare performance wise with using the listSecretVersionIdsPaginator approach? Looks like this will force all the retrieval of the secrets metadata upfront before retrieval the secret values themselves.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +80,39 @@ public Optional<String> fetchSecret(final String secretName) {
...
+
+  public <R> Collection<R> mapSecrets(
+      final AbstractMap<String, String> tags, final BiFunction<String, String, R> mapper) {
+    final Set<R> result = ConcurrentHashMap.newKeySet();
+    listSecrets(tags)
+        .parallelStream()
+        .forEach(
+            secretEntry -> {
+              try {
+                final String secretValue = fetchSecret(secretEntry.name()).get();
```
jframe, line 102: This doesn't handle the case where fetchSecret returns an Optional.empty

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +81,54 @@ public Optional<String> fetchSecret(final String secretName) {
...
+        .iterator()
+        .forEachRemaining(
+            listSecretsResponse -> {
+              listSecretsResponse
+                  .secretList()
+                  .parallelStream()
+                  .forEach(
+                      secretEntry -> {
+                        try {
+                          final String secretValue = fetchSecret(secretEntry.name()).get();
```
jframe, line 113: This doesn't handle the case where a secret isn't found. The mapper should only try and map the value if it exists. Not that likely but it's possible that the secret no longer exists when we go to fetch the value. In the case we get an empty result I think we can just log a warning and skip it like we do with other errors.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/AwsSecretsManager.java
```diff
@@ -68,6 +81,54 @@ public Optional<String> fetchSecret(final String secretName) {
...
+    }
+    if (!tagValues.isEmpty()) {
+      listSecretsRequestBuilder.filters(
+          Filter.builder().key(FilterNameStringType.TAG_VALUE).values(tagValues).build());
+    }
+    return secretsManagerClient.listSecretsPaginator(listSecretsRequestBuilder.build());
+  }
+
+  public <R> Collection<R> mapSecrets(
+      final List<String> tagKeys,
```
jframe, line 99: Does this need to be a List? The filter takes a collection so there probably no need to this be a list, a collection is probably sufficient.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -101,46 +104,222 @@ private void closeClients() {
...
+    secretNames.forEach(secretName -> validateMappedSecret(secretEntries, secretName));
+  }
+
+  @Test
+  void listAndMapMultipleSecretsWithMultipleTags() {
+    createSecret(true, true, false);
+
+    final Collection<AbstractMap.SimpleEntry<String, String>> secretEntries =
+        awsSecretsManagerExplicit.mapSecrets(
+            secretTags.keySet().stream().collect(Collectors.toList()),
```
jframe, line 244: nit: this can be simpler now that collection is being used, can use secretTags.keySet() and same for the other tests

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -101,46 +104,222 @@ private void closeClients() {
...
+            secretTags.values().stream().collect(Collectors.toList()),
```
jframe, line 245: nit: similarly this can be simpler now that collection is being used, can use secretTags.values() and same for the other tests

jframe (approved):

PR #146, opened 2022-05-04
Fixed aws test failures + refactoring

georgep9:
Fixed keystorage aws test failures + refactoring. AWS Secrets Managers are now initialized and closed, before and after each test case respectively to avoid race conditions when creating and deleting secrets.

Related to https://github.com/ConsenSys/web3signer/issues/499

### DIFF
--- a/keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
+++ b/keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
@@ -22,45 +22,47 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Collectors;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInstance;
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 import software.amazon.awssdk.services.secretsmanager.model.CreateSecretRequest;
+import software.amazon.awssdk.services.secretsmanager.model.CreateSecretResponse;
 import software.amazon.awssdk.services.secretsmanager.model.DeleteSecretRequest;
 import software.amazon.awssdk.services.secretsmanager.model.Tag;
 
-@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class AwsSecretsManagerTest {
 
-  private final String RW_AWS_ACCESS_KEY_ID = System.getenv("RW_AWS_ACCESS_KEY_ID");
-  private final String RW_AWS_SECRET_ACCESS_KEY = System.getenv("RW_AWS_SECRET_ACCESS_KEY");
+  private static final String RW_AWS_ACCESS_KEY_ID = System.getenv("RW_AWS_ACCESS_KEY_ID");
+  private static final String RW_AWS_SECRET_ACCESS_KEY = System.getenv("RW_AWS_SECRET_ACCESS_KEY");
 
-  private final String RO_AWS_ACCESS_KEY_ID = System.getenv("RO_AWS_ACCESS_KEY_ID");
-  private final String RO_AWS_SECRET_ACCESS_KEY = System.getenv("RO_AWS_SECRET_ACCESS_KEY");
-  private final String AWS_REGION = "us-east-2";
+  private static final String RO_AWS_ACCESS_KEY_ID = System.getenv("RO_AWS_ACCESS_KEY_ID");
+  private static final String RO_AWS_SECRET_ACCESS_KEY = System.getenv("RO_AWS_SECRET_ACCESS_KEY");
+  private static final String AWS_REGION = "us-east-2";
 
-  private AwsSecretsManager awsSecretsManagerDefault;
-  private AwsSecretsManager awsSecretsManagerExplicit;
-  private AwsSecretsManager awsSecretsManagerInvalidCredentials;
-  private SecretsManagerClient secretsManagerClient;
-  private String secretName;
-  private String secretNamePrefix;
-  private List<String> secretNames;
-  private AbstractMap<String, String> secretTags;
+  private static AwsSecretsManager awsSecretsManagerDefault;
+  private static AwsSecretsManager awsSecretsManagerExplicit;
+  private static AwsSecretsManager awsSecretsManagerInvalidCredentials;
+  private static AwsBasicCredentials awsBasicCredentials;
+  private static StaticCredentialsProvider credentialsProvider;
+  private static SecretsManagerClient testSecretsManagerClient;
+  private static String testSecretName;
+  private static String testSecretNamePrefix;
+  private static List<String> testSecretNames;
+  private static AbstractMap<String, String> testSecretTags;
 
-  private static final String SECRET_VALUE =
+  private final String SECRET_VALUE =
       "{\"crypto\": {\"kdf\": {\"function\": \"scrypt\", \"params\": {\"dklen\": 32, \"n\": 262144, \"r\": 8, \"p\": 1, \"salt\": \"3d9b30b612f4f5e9423dc43c0490396798a179d35dd58d48dc1f5d6d42b07ab6\"}, \"message\": \"\"}, \"checksum\": {\"function\": \"sha256\", \"params\": {}, \"message\": \"c762b7453eab3332cda31d9dee1894cf541373617e591a8e7ab8f14f5830f723\"}, \"cipher\": {\"function\": \"aes-128-ctr\", \"params\": {\"iv\": \"095f79f6bb5daab60355ab6aa894b3c8\"}, \"message\": \"4ca342a769ec1c00d6a6d69e18cdf821f42849d4431da7df827b01ba162ed763\"}}, \"description\": \"\", \"pubkey\": \"8fb7c68f3291b8db46ef86a8b9544cad7052dd7cf817862063d1f151f3c443cd3907830b09a86fe0513f0e863beccf25\", \"path\": \"m/12381/3600/0/0/0\", \"uuid\": \"88fc9701-8670-4378-a3ba-00be25c1330c\", \"version\": 4}";
 
-  private void verifyEnvironmentVariables() {
+  @BeforeAll
+  static void verifyEnvironmentVariables() {
     Assumptions.assumeTrue(
         RW_AWS_ACCESS_KEY_ID != null, "Set RW_AWS_ACCESS_KEY_ID environment variable");
     Assumptions.assumeTrue(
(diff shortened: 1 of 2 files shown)


keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +73,287 @@ private void verifyEnvironmentVariables() {
...
   @Test
   void fetchSecretWithDefaultManager() {
-    createSecret(false, false, false);
-    Optional<String> secret = awsSecretsManagerDefault.fetchSecret(secretName);
+    createSecret(false, false);
+
+    Optional<String> secret = awsSecretsManagerDefault.fetchSecret(testSecretName);
     assertThat(secret).hasValue(SECRET_VALUE);
+
+    deleteSecrets();
```
jframe, line 110: Can this just be done in a @AfterEach to clean up?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +73,287 @@ private void verifyEnvironmentVariables() {
...
+  }
+
+  private CreateSecretResponse createSecret(final boolean multipleTags, final boolean sharedTag) {
+    testSecretNamePrefix = "signers-aws-integration/";
+    testSecretName = testSecretNamePrefix + UUID.randomUUID();
+    testSecretNames.add(testSecretName);
+
+    final List<Tag> tags = new ArrayList<>();
+
+    if (sharedTag) {
```
jframe, line 305: Think it would be clearer to just pass in any tags to associated with the secrets

jframe, line 305: Think it would be clearer to just pass in any tags to associated with the secrets

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -22,45 +22,47 @@
...
+  private static AwsSecretsManager awsSecretsManagerInvalidCredentials;
+  private static AwsBasicCredentials awsBasicCredentials;
+  private static StaticCredentialsProvider credentialsProvider;
+  private static SecretsManagerClient testSecretsManagerClient;
+  private static String testSecretName;
+  private static String testSecretNamePrefix;
+  private static List<String> testSecretNames;
+  private static AbstractMap<String, String> testSecretTags;
+
+  private final String SECRET_VALUE =
```
jframe, line 61: nit: we don't care about the value so long as it returns the right one so could just make this a lot simpler with a smaller test value

jframe, line 61: nit: we don't care about the value so long as it returns the right one so could just make this a lot simpler with a smaller test value

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +73,287 @@ private void verifyEnvironmentVariables() {
...
+  @AfterEach
+  void closeTestClients() {
+    closeAwsSecretsManagers();
+  }
+
   @Test
   void fetchSecretWithDefaultManager() {
-    createSecret(false, false, false);
-    Optional<String> secret = awsSecretsManagerDefault.fetchSecret(secretName);
+    createSecret(false, false);
```
jframe, line 105: Is there any benefit to creating the secrets for each test? Could we just create them all in a beforeAll so simplify the tests?

jframe (commented):
Running this locally the test is quite intermittent. Think the approach of creating before each test might be problematic. Might be better off just creating them before all the tests run or even just have them already created in the secrets manager like we do with AWS testing.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +76,294 @@ private void verifyEnvironmentVariables() {
...
+        AwsBasicCredentials.create(RW_AWS_ACCESS_KEY_ID, RW_AWS_SECRET_ACCESS_KEY);
+    credentialsProvider = StaticCredentialsProvider.create(awsBasicCredentials);
+    testSecretsManagerClient =
+        SecretsManagerAsyncClient.builder()
+            .credentialsProvider(credentialsProvider)
+            .region(Region.of(AWS_REGION))
+            .build();
+  }
+
+  private static void initTestVariables() {
```
jframe, line 249: Could this initialisation just be done in the field declaration?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +76,294 @@ private void verifyEnvironmentVariables() {
...
+    testSecretMultipleSharedTags.clear();
+  }
+
+  private static Tag createTag(final String key, final String value) {
+    return Tag.builder().key(key).value(value).build();
+  }
+
+  private static void updateTestTags(
+      final List<Tag> tags, final boolean multipleTags, final boolean sharedTags) {
+    tags.forEach(
```
jframe, line 330: nit: find this a bit confusing. Think it would be easier to follow if we just created the sets of tags upfront and then pass them to the secret we want to be created instead.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -25,42 +25,47 @@
...
+  private static AwsSecretsManager awsSecretsManagerDefault;
+  private static AwsSecretsManager awsSecretsManagerExplicit;
+  private static AwsSecretsManager awsSecretsManagerInvalidCredentials;
+  private static AwsBasicCredentials awsBasicCredentials;
+  private static StaticCredentialsProvider credentialsProvider;
+  private static SecretsManagerAsyncClient testSecretsManagerClient;
+  private static String testSecretName;
+  private static String testSecretNamePrefix;
+  private static List<String> testSecretNames;
+  private static AbstractMap<String, String> allTestSecretTags;
```
jframe, line 58: This can be just Map<String, String>

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +76,294 @@ private void verifyEnvironmentVariables() {
...
-
   @BeforeAll
-  void setup() {
-    verifyEnvironmentVariables();
-    setupSecretsManagers();
-    setupSecretsManagerClient();
-    initializeVariables();
+  static void setup() {
+    initAwsSecretsManagers();
+    initTestSecretsManager();
```
siladu, line 82: Would initTestSecretsMangerClient be a more accurate name?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -71,255 +76,298 @@ private void verifyEnvironmentVariables() {
...
+    testTags.keySet().stream()
+        .filter(tagKey -> testSecretNames.contains(tagKey))
+        .collect(Collectors.toList())
+        .forEach(
+            tagKey ->
+                assertThat(tagKey)
+                    .isIn(
+                        secretEntries.stream()
+                            .map(entry -> entry.getKey())
+                            .collect(Collectors.toList())));
```
siladu, line 371:
```suggestion
    testTags.keySet().stream()
        .filter(tagKey -> testSecretNames.contains(tagKey))
        .forEach(tagKey ->
            assertThat(tagKey).isIn(secretEntries.stream().map(e -> e.getKey()))
        );
```

This would remove the unnecessary list conversions.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
siladu: Agree it would be easier to set up some tags first, then then match them up with the secrets as required.

siladu (changes requested):
I am having trouble understanding the testSecret* maps. Rather than having to mentally keep track of multiple bits of shared and interacting state, it would be easier to understand if we explicitly declared some tags and associated them with secrets as mentioned in the other comment.

I'm not sure the permutations of single/multiple/shared tags are precisely covering the logic within AwsSecretsManager. For the listAndMap... tests, I would aim to cover the four code paths inside listSecrets, i.e. no filters, just tagKeys, just tagValues, both tagKeys and tagValues.

Should passing empty filters result in returning all secrets or no secrets?

siladu (commented):
made a couple of suggestions

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +71,216 @@ private void verifyEnvironmentVariables() {
...
+  // secret1: tag(k1, vA)
+  // secret2: tag(k1, vB)
+  // secret3: tag(k2, vC)
+  // secret4: tag(k2, vB)
+  private static void createTestSecrets() {
+    secretName1 = createTestSecret("tagKey1", "tagValA");
+    secretName2 = createTestSecret("tagKey1", "tagValB");
+    secretName3 = createTestSecret("tagKey2", "tagValC");
+    secretName4 = createTestSecret("tagKey2", "tagValB");
+    testSecretNames.addAll(List.of(secretName1, secretName2, secretName3, secretName4));
```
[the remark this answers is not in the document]
georgep9: might be tricky if we need to initlialize the clients beforehand. Having the stages separated into functions for the setup() routine reads and flows better IMO.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +73,208 @@ private void verifyEnvironmentVariables() {
...
-    secretNames.forEach(secretName -> validateMappedSecret(secretEntries, secretName));
-  }
-
-  @Test
-  void throwsAwayObjectsThatFailMapper() {
-    createSecret(true, false, false);
-
-    final String failEntryName = secretNames.get(1);
-
+  void throwsAwayObjectsWhichMapToNull() {
```
jframe, line 183: What happens if the mapper throws an exception? Can we have a test to show that the AwsSecretsManager handles that

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +73,208 @@ private void verifyEnvironmentVariables() {
...
-            AbstractMap.SimpleEntry::new);
-
-    secretNames.forEach(secretName -> validateMappedSecret(secretEntries, secretName));
+            List.of("tagKey1"), Collections.emptyList(), AbstractMap.SimpleEntry::new);
+
+    final List<String> secretNames =
+        secretEntries.stream()
+            .map(entry -> String.valueOf(entry.getKey()))
+            .collect(Collectors.toList());
+    assertThat(secretNames)
```
jframe, line 144: This doesn't prove that we actually fetched any secrets or applied the mapping function. Can we check that the results can the expected secrets as well?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +73,208 @@ private void verifyEnvironmentVariables() {
...
+    AwsBasicCredentials awsBasicCredentials =
+        AwsBasicCredentials.create(RW_AWS_ACCESS_KEY_ID, RW_AWS_SECRET_ACCESS_KEY);
+    StaticCredentialsProvider credentialsProvider =
+        StaticCredentialsProvider.create(awsBasicCredentials);
+    testSecretsManagerClient =
+        SecretsManagerAsyncClient.builder()
+            .credentialsProvider(credentialsProvider)
+            .region(Region.of(AWS_REGION))
+            .build();
+  }
```
jframe, line 219: Can a test also be added for the case when no tags are matched

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +73,208 @@ private void verifyEnvironmentVariables() {
...
+    return testSecretName;
+  }
+
+  private String createTestSecret(final String tagKey, final String tagVal) {
+    final Tag testSecretTag = Tag.builder().key(tagKey).value(tagVal).build();
+    return createSecret(testSecretTag);
+  }
+
+  private void createTestSecrets() {
+    secretName1 = createTestSecret("tagKey1", "tagValA");
```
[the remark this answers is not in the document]
georgep9: Arbitrary values are chosen here for certain test cases

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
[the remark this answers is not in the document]
siladu: ah yes, I forgot they rely on the client initialisation too

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -50,19 +48,21 @@ class AwsSecretsManagerTest {
   private final String RO_AWS_SECRET_ACCESS_KEY = System.getenv("RO_AWS_SECRET_ACCESS_KEY");
   private final String AWS_REGION = "us-east-2";
 
+  private final String SECRET_NAME_PREFIX = "signers-aws-integration/";
+  private final String SECRET_VALUE = "secret-value";
```
siladu, line 52: nit: constants could be static

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +76,242 @@ private void verifyEnvironmentVariables() {
...
+            .map(entry -> String.valueOf(entry.getKey()))
+            .collect(Collectors.toList());
+    final List<String> secretValues =
+        secretEntries.stream()
+            .map(entry -> String.valueOf(entry.getValue()))
+            .collect(Collectors.toList());
+    assertThat(secretNames)
+        .contains(secretName1WithTagKey1ValA, secretName2WithTagKey1ValB)
+        .doesNotContain(secretName3WithTagKey2ValC, secretName4WithTagKey2ValB);
+    assertThat(secretValues).contains(SECRET_VALUE);
```
siladu, line 173: nit: If we're only checking the mapped value in one test, then I think it makes more sense to either
1. rename the other tests that don't check this from listAndMap... to list...
2. Create a separate test just to assert the map functionality.

Alternatively you could assert the values on the other tests as well and then "listAndMap..." would be accurate

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -73,255 +76,242 @@ private void verifyEnvironmentVariables() {
...
+    closeAwsSecretsManagers();
+    closeTestSecretsManager();
+  }
 
-    final Optional<AbstractMap.SimpleEntry<String, String>> nullEntry =
-        secretEntries.stream().filter(e -> e.getKey().equals("MyBls")).findAny();
-    assertThat(nullEntry).isEmpty();
+  private String createSecret(final Tag tag)
+      throws ExecutionException, InterruptedException, TimeoutException {
+    final String testSecretName = SECRET_NAME_PREFIX + "/" + UUID.randomUUID();
```
siladu, line 265: nit: spurious "/"

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -175,25 +206,26 @@ void listAndMapSecretsWithMatchingTagKeysAndValues() {
...
-              if (name.equals(secretName1)) {
+              if (name.equals(secretName1WithTagKey1ValA)) {
                 return null;
               }
               return new AbstractMap.SimpleEntry<>(name, value);
             });
 
     final Optional<AbstractMap.SimpleEntry<String, String>> nullEntry =
-        secretEntries.stream().filter(e -> e.getKey().equals(secretName1)).findAny();
+        secretEntries.stream().filter(e -> e.getKey().equals(secretName1WithTagKey1ValA)).findAny();
```
siladu, line 223: I think line 227-229 equivalent to:
```java
assertThat(secretEntries.stream().map(e -> e.getKey())).doesNotContain(secretName1WithTagKey1ValA);
```

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -127,7 +132,27 @@ void emptyTagFiltersReturnAllSecrets() {
...
+  void nonExistentTagFiltersReturnsEmpty() {
+    final Collection<AbstractMap.SimpleEntry<String, String>> secretEntries =
+        awsSecretsManagerExplicit.mapSecrets(
+            List.of("nonexistent-tag-key"),
+            List.of("nonexistent-tag-value"),
+            AbstractMap.SimpleEntry::new);
+
+    final List<String> secretNames =
+        secretEntries.stream()
+            .map(entry -> String.valueOf(entry.getKey()))
```
siladu, line 151: I think all these String.valueOf are redundant

siladu (commented):
Some inline comments. Also, it's not clear to me if and how the scenario where the mapper throws an exception is covered.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
jframe, line 173: With these all having the same secret value we aren't testing that all the secret values are returned. I think the secret values need to be unique for these to be a valid test.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
jframe: Wasn't as clear about this as I should have been. I meant if the user supplied mapper function to mapSecrets fails for any reason we should handle that and continue process other secrets with the mapping function.

siladu (approved):
LGTM!

jframe (approved):

PR #147, opened 2022-05-06
Include AWS STS library so that web identity tokens can be used

jframe:
Include AWS STS library so that web identity tokens can be used. It's not on the class path by default.

There is a note here that the STS library is required on the classpath https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/auth/credentials/WebIdentityTokenFileCredentialsProvider.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.2.2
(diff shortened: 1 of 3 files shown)


usmansaleem (approved):
LGTM

PR #150, opened 2022-05-17
Keystorage - aws test secrets fix

georgep9:
Updated createSecret to rely on existing identical tests secrets. This is to avoid flaky test scenarios when managing test secrets every run. The deleteSecrets routine was also removed.

Supports issue ConsenSys/web3signer#499.

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -23,7 +23,7 @@ dependencyManagement {
     }
     dependency 'com.google.guava:guava:31.1-jre'
 
-    dependencySet(group: 'com.azure', version: '4.3.8') {
+    dependencySet(group: 'com.azure', version: '4.4.2') {
       entry 'azure-security-keyvault-secrets'
       entry 'azure-security-keyvault-keys'
     }
@@ -35,7 +35,7 @@ dependencyManagement {
 
     dependency 'io.rest-assured:rest-assured:4.4.0'
 
-    dependencySet(group: 'io.vertx', version: '4.2.6') {
+    dependencySet(group: 'io.vertx', version: '4.2.7') {
       entry 'vertx-codegen'
       entry 'vertx-core'
       entry 'vertx-unit'
@@ -94,6 +94,19 @@ dependencyManagement {
       entry 'auth'
       entry 'secretsmanager'
       entry 'sts'
+    }
+
+    // explicit declaring to override transitive dependencies with vulnerabilities
+    // addressing CVE-2022-24823
+    dependencySet(group: 'io.netty', version: '4.1.77.Final') {
+      entry 'netty-codec-http2'
+      entry 'netty-handler-proxy'
+      entry 'netty-resolver-dns'
+      entry 'netty-resolver-dns-native-macos'
+      entry 'netty-transport-native-epoll'
+      entry 'netty-transport-native-kqueue'
+      entry 'netty-transport-native-unix-common'
+      entry 'netty-transport-classes-epoll'
     }
   }
 }
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -95,5 +95,7 @@ dependencyManagement {
       entry 'secretsmanager'
       entry 'sts'
     }
+
+    dependency 'org.awaitility:awaitility:4.2.0'
   }
 }
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -95,7 +95,5 @@ dependencyManagement {
       entry 'secretsmanager'
       entry 'sts'
     }
-
-    dependency 'org.awaitility:awaitility:4.2.0'
   }
 }
--- a/keystorage/aws/build.gradle
+++ b/keystorage/aws/build.gradle
@@ -42,7 +42,6 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-engine'
   testImplementation 'org.junit.jupiter:junit-jupiter-api'
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
-  testImplementation 'org.awaitility:awaitility'
 
   errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/aws/build.gradle
+++ b/keystorage/aws/build.gradle
@@ -42,6 +42,7 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-engine'
   testImplementation 'org.junit.jupiter:junit-jupiter-api'
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
+  testImplementation 'org.awaitility:awaitility'
 
   errorprone("com.google.errorprone:error_prone_core")
 }
--- a/keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
+++ b/keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
@@ -278,10 +278,9 @@ class AwsSecretsManagerTest {
               DescribeSecretRequest.builder().secretId(testSecretName).build());
       final boolean hasDifferentSecretValue =
           !getSecretValueResponse.secretString().equals(secretValue);
-      final boolean hasDifferentSecretTag =
-          describeSecretResponse.hasTags()
-              && !describeSecretResponse.tags().get(0).equals(testSecretTag);
-      if (hasDifferentSecretValue || hasDifferentSecretTag) {
+      final boolean doesNotHaveTags = !describeSecretResponse.hasTags();
+      final boolean hasDifferentSecretTag = !describeSecretResponse.tags().contains(testSecretTag);
+      if (hasDifferentSecretValue || doesNotHaveTags || hasDifferentSecretTag) {
         updateSecret(
             testSecretName, secretValue, describeSecretResponse.tags().get(0), testSecretTag);
       }
--- a/keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
+++ b/keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
@@ -21,7 +21,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
-import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -36,8 +35,11 @@ import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.secretsmanager.SecretsManagerAsyncClient;
 import software.amazon.awssdk.services.secretsmanager.model.CreateSecretRequest;
-import software.amazon.awssdk.services.secretsmanager.model.DeleteSecretRequest;
+import software.amazon.awssdk.services.secretsmanager.model.DescribeSecretRequest;
+import software.amazon.awssdk.services.secretsmanager.model.DescribeSecretResponse;
 import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueRequest;
+import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
+import software.amazon.awssdk.services.secretsmanager.model.ResourceNotFoundException;
 import software.amazon.awssdk.services.secretsmanager.model.Tag;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
@@ -87,11 +89,10 @@ class AwsSecretsManagerTest {
   }
 
   @AfterAll
-  void teardown() throws Exception {
+  void teardown() {
     if (awsSecretsManagerDefault != null
         || awsSecretsManagerExplicit != null
         || testSecretsManagerClient != null) {
-      deleteTestSecrets();
       closeClients();
     }
   }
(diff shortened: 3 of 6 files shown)


keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -265,9 +263,34 @@ private void closeClients() {
...
+              .get(30, TimeUnit.SECONDS);
+      final DescribeSecretResponse describeSecretResponse =
+          testSecretsManagerClient
+              .describeSecret(DescribeSecretRequest.builder().secretId(testSecretName).build())
+              .get(30, TimeUnit.SECONDS);
+      return getSecretValueResponse.name().equals(testSecretName)
+          && getSecretValueResponse.secretString().equals(secretValue)
+          && describeSecretResponse.hasTags()
+          && describeSecretResponse.tags().get(0).key().equals(tag.key())
+          && describeSecretResponse.tags().get(0).value().equals(tag.value());
```
siladu, line 281:
```suggestion
return getSecretValueResponse.secretString().equals(secretValue)
          && describeSecretResponse.tags().contains(tag);
```

You used the testSecretName to get fetch the details so I think it's redundant to check its equality here too.
Can also use contains and rely on Tag's .equals()

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -265,9 +263,34 @@ private void closeClients() {
     closeTestSecretsManager();
   }
 
-  private String createSecret(final Tag tag, String secretValue)
+  private boolean checkTestSecretExists(
```
siladu, line 266: It's a shame we have to spend two requests getting this info. I can see how it's useful to be able to update the secrets in storage just by changing the tests though. 

In the scenario where we update either the secret values or tags, I think this would leave orphaned keys...do you think we should delete any matching key names whose values/tags don't match as part of this too?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -282,27 +305,20 @@ private String createSecret(final Tag tag, String secretValue)
   }
 
   private String createTestSecret(
-      final String tagKey, final String tagVal, final String secretValue) throws Exception {
-    final Tag testSecretTag = Tag.builder().key(tagKey).value(tagVal).build();
+      final String secretName, final String tagKey, final String tagVal, final String secretValue)
+      throws Exception {
```
siladu, line 309: Since this method can now throw an Exception, I think you can remove the try catch in this method and just let createSecret's Exception bubble up

siladu (commented):
I like the create or update approach, good idea!

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -265,9 +263,34 @@ private void closeClients() {
...
+  private boolean checkTestSecretExists(
+      final String testSecretName, final Tag tag, final String secretValue) {
+    try {
+      final GetSecretValueResponse getSecretValueResponse =
+          testSecretsManagerClient
+              .getSecretValue(GetSecretValueRequest.builder().secretId(testSecretName).build())
+              .get(30, TimeUnit.SECONDS);
+      final DescribeSecretResponse describeSecretResponse =
+          testSecretsManagerClient
+              .describeSecret(DescribeSecretRequest.builder().secretId(testSecretName).build())
```
jframe, line 275: Is it necessary to check the tags or the secret? These are secrets are already setup so only need check the secret itself exists.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -282,27 +305,20 @@ private String createSecret(final Tag tag, String secretValue)
...
-    secretName3WithTagKey2ValC = createTestSecret("tagKey2", "tagValC", SECRET_VALUE3);
-    secretName4WithTagKey2ValB = createTestSecret("tagKey2", "tagValB", SECRET_VALUE4);
-    testSecretNames = new ArrayList<>();
-    testSecretNames.addAll(
-        List.of(
-            secretName1WithTagKey1ValA,
-            secretName2WithTagKey1ValB,
-            secretName3WithTagKey2ValC,
-            secretName4WithTagKey2ValB));
+    secretName1WithTagKey1ValA = createTestSecret("secret1", "tagKey1", "tagValA", SECRET_VALUE1);
```
jframe, line 318: Be good to make the secret names constant now since that don't change.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
jframe, line 309: Think throwing a runtime exception would be more appropriate in a unit test

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -265,9 +263,34 @@ private void closeClients() {
...
+    try {
+      final GetSecretValueResponse getSecretValueResponse =
+          testSecretsManagerClient
+              .getSecretValue(GetSecretValueRequest.builder().secretId(testSecretName).build())
+              .get(30, TimeUnit.SECONDS);
+      final DescribeSecretResponse describeSecretResponse =
+          testSecretsManagerClient
+              .describeSecret(DescribeSecretRequest.builder().secretId(testSecretName).build())
+              .get(30, TimeUnit.SECONDS);
+      return getSecretValueResponse.name().equals(testSecretName)
```
jframe, line 277: If only the tag or value changed will create secret change that? or will it fail creating the secrets before tag changes are applied?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -260,14 +270,34 @@ private void closeAwsSecretsManagers() {
...
+    secretName2WithTagKey1ValB =
+        createTestSecret(SECRET_NAME2, "tagKey1", "tagValB", SECRET_VALUE2);
+    secretName3WithTagKey2ValC =
+        createTestSecret(SECRET_NAME3, "tagKey2", "tagValC", SECRET_VALUE3);
+    secretName4WithTagKey2ValB =
+        createTestSecret(SECRET_NAME4, "tagKey2", "tagValB", SECRET_VALUE4);
   }
 
-  private String createSecret(final Tag tag, String secretValue)
+  private String createTestSecret(
```
siladu, line 284: All this message is really doing now is create the tag...might be neater to move Tag create up one level and delete this method?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -281,28 +311,49 @@ private String createSecret(final Tag tag, String secretValue)
...
+              .get(30, TimeUnit.SECONDS);
+      if (!getSecretValueResponse.secretString().equals(secretValue)
+          && describeSecretResponse.hasTags()
+          && !describeSecretResponse.tags().get(0).equals(tag)) {
+        updateSecret(testSecretName, secretValue, describeSecretResponse.tags().get(0), tag);
+      }
+      return true;
     } catch (Exception e) {
-      throw new Exception(e.getMessage());
+      return false;
```
siladu, line 332: If the secret exists but we throw an exception while checking and/or updating it then I think we should throw an exception here rather than return false. 

Returning false will result in a new CreateSecretRequest which would fail in this scenario.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -281,28 +311,49 @@ private String createSecret(final Tag tag, String secretValue)
...
   }
 
-  private String createTestSecret(
-      final String tagKey, final String tagVal, final String secretValue) throws Exception {
-    final Tag testSecretTag = Tag.builder().key(tagKey).value(tagVal).build();
+  private boolean checkTestSecretExistsAndUpdateIfUnmatched(
+      final String testSecretName, final Tag tag, final String secretValue) {
     try {
-      return createSecret(testSecretTag, secretValue);
+      final GetSecretValueResponse getSecretValueResponse =
```
siladu, line 317: How is a non-existent secret represented in this response? I suggest we inspect for that and return false before continuing with the code below

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
georgep9: Would it hurt to bottle up in that fashion? Don't think we care to debug when the secret fails to update, only when creating, listing, and fetching.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
georgep9: The request is handled exceptionally so whenever these fail, we know the secret is either non-existent or broken since we're failing to fetch and describe.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
siladu: If the secret fails to update then this code would return false and therefore attempt to recreate the same secret, which would fail because it already exists.

I think we can only continue with the test if all attempted AWS calls (including updating) are successful, so in my mind it makes sense to always bubble up AWS exceptions.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
siladu: So there's no way of telling the difference between "exception occurred during request" and "valid request but secret doesn't exist"?

georgep9: Yep, you need to handle that exceptionally with ResourceNotFoundException for example

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
georgep9: in that case the test fails to run. We can then try again or manually inspect with the console

georgep9: I guess not ideal on the first pipeline run

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
siladu: In that case I think we only want to return false in the case of ResourceNotFoundException, all other Exceptions should be thrown

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -260,64 +269,99 @@ private void closeAwsSecretsManagers() {
...
+      final String testSecretName, final String secretValue, final Tag oldTag, final Tag newTag)
+      throws RuntimeException {
+    try {
+      testSecretsManagerClient
+          .updateSecret(
+              UpdateSecretRequest.builder()
+                  .secretId(testSecretName)
+                  .secretString(secretValue)
+                  .build())
+          .get(30, TimeUnit.SECONDS);
```
siladu, line 345: Is there a delay between updating the secret and the new value being reflected in the GetSecretValueResponse?

Wondering if we'd need something like waitUntilSecretAvailable here...hopefully not, but might be worth considering if this turns out to be flaky again.

gradle/versions.gradle
```diff
@@ -95,5 +95,7 @@ dependencyManagement {
       entry 'secretsmanager'
       entry 'sts'
     }
+
+    dependency 'org.awaitility:awaitility:4.2.0'
```
siladu, line 99: Is this still needed?

keystorage/aws/build.gradle
```diff
@@ -42,6 +42,7 @@ dependencies {
   testImplementation 'org.junit.jupiter:junit-jupiter-engine'
   testImplementation 'org.junit.jupiter:junit-jupiter-api'
   testImplementation 'org.junit.jupiter:junit-jupiter-params'
+  testImplementation 'org.awaitility:awaitility'
```
siladu, line 45: Is this still needed?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/AwsSecretsManagerTest.java
```diff
@@ -260,64 +256,53 @@ private void closeAwsSecretsManagers() {
...
+              GetSecretValueRequest.builder().secretId(testSecretName).build());
+      final DescribeSecretResponse describeSecretResponse =
+          testSecretsManagerClient.describeSecret(
+              DescribeSecretRequest.builder().secretId(testSecretName).build());
+      final boolean hasDifferentSecretValue =
+          !getSecretValueResponse.secretString().equals(secretValue);
+      final boolean hasDifferentSecretTag =
+          describeSecretResponse.hasTags()
+              && !describeSecretResponse.tags().get(0).equals(testSecretTag);
+      if (hasDifferentSecretValue || hasDifferentSecretTag) {
```
siladu, line 284: Think it might need to be: `hasDifferentSecretValue || hasNoTags || hasDifferentSecretTag`
to cover the case where a tag is removed, e.g. manually.

siladu (approved):
LGTM

PR #151, opened 2022-05-18
Add gradle dependency check plugin

siladu:
Same config as ethsigner and web3signer.

Fix CVE-2022-24823 by locking in netty version 4.1.77

Couple of relevant upgrades.


jframe (approved):
LGTM

PR #152, opened 2022-05-23
AWS Secrets Manager mapSecrets with prefixes

georgep9:
Implemented secret name prefixes in mapSecrets. The name prefix is applied during the list secrets request as a Filter, following the same approach used for tagKeys and tagValues. This builds upon the secrets manager search interface referenced in the [Find secrets in AWS Secrets Manager](https://docs.aws.amazon.com/secretsmanager/latest/userguide/manage_search-secret.html) user guide.

Supports [ConsenSys/websigner#499](https://github.com/ConsenSys/web3signer/issues/499)


usmansaleem:
Will be applied via #154

PR #164, opened 2022-12-12
Various dependencies versions update

usmansaleem:
- Update various dependencies version
  - Update dependency checker gradle plugin version
  - Update Tuweni version
  - Update jackson version
  - Update vertx version
  - Update errorprone version
- Add missing dependency in acceptance test which was previously resolved via transitive dependency


siladu (approved):
LGTM

PR #166, opened 2023-01-04
Update netty to address CVE-2022-41881 and CVE-2022-41915

jframe:
Update netty to address CVE-2022-41881 and CVE-2022-41915


siladu (approved):
LGTM

PR #168, opened 2023-01-05
Update AWS acceptance tests to use RO prefix to RO AWS creds

jframe:


jframe:
No longer needed as builder environment variables have been updated

PR #172, opened 2023-01-06
Various dependency upgrades

siladu:
Disable a couple of tests which don't make sense after upgrading beyond vertx 4.3.4. These will be rewritten in a subsequent commit.

Hashicorp support for mac m1 arch in acceptance test


jframe (approved):
LGTM

PR #173, opened 2023-01-06
Fix hashicorp connection tests disabled in the previous commit

siladu:
Move tests from factory to the connection itself since vertx 4.3.4 moved the TLS loading code from connection creation to the first request (fetchKey)


jframe (approved):
LGTM

PR #175, opened 2023-02-06
AwsSecretsManager allows loading multiple secrets from same property

usmansaleem:
Allows loading keys separated by line terminators from the same property in AWS Secrets Manager.

Required for https://github.com/ConsenSys/web3signer/issues/706

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -137,5 +137,8 @@ dependencyManagement {
        \-- org.apache.tuweni:tuweni-net:2.3.1
      */
     dependency 'commons-net:commons-net:3.9.0'
+
+    // used in tests to assert log message
+    dependency 'de.neuland-bfi:assertj-logging-log4j:0.5.0'
   }
 }
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2023 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.signers.aws;
+
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class SecretValueConverterUtil {
+  private static final Logger LOG = LogManager.getLogger();
+
+  public static <R> Set<R> mapSecretValue(
+      BiFunction<String, String, R> mapper, String propKey, String secretValue) {
+    final AtomicLong discardedValues = new AtomicLong(0);
+    final Set<R> mappedValues =
+        secretValue
+            .lines()
+            .map(
+                v -> {
+                  final R obj = mapper.apply(propKey, v);
+                  if (obj == null) {
+                    discardedValues.incrementAndGet();
+                  }
+                  return obj;
+                })
+            .filter(Objects::nonNull)
+            .collect(Collectors.toSet());
+
+    if (discardedValues.get() > 0) {
+      LOG.warn(
+          "AWS Secrets - {} values from property {} were not mapped and discarded.",
+          discardedValues.get(),
+          propKey);
+    }
+    return mappedValues;
+  }
+}
--- a/keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
+++ b/keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
@@ -26,27 +26,22 @@ public class SecretValueConverterUtil {
 
   public static <R> Set<R> mapSecretValue(
       BiFunction<String, String, R> mapper, String propKey, String secretValue) {
-    final AtomicLong discardedValues = new AtomicLong(0);
+    final AtomicLong valuesIndex = new AtomicLong(0);
     final Set<R> mappedValues =
         secretValue
             .lines()
             .map(
                 v -> {
+                  long index = valuesIndex.getAndIncrement();
                   final R obj = mapper.apply(propKey, v);
                   if (obj == null) {
-                    discardedValues.incrementAndGet();
+                    LOG.warn("Secret Value {}:{} was not mapped and discarded.", propKey, index);
                   }
                   return obj;
                 })
             .filter(Objects::nonNull)
             .collect(Collectors.toSet());
 
-    if (discardedValues.get() > 0) {
-      LOG.warn(
-          "AWS Secrets - {} values from property {} were not mapped and discarded.",
-          discardedValues.get(),
-          propKey);
-    }
     return mappedValues;
   }
 }
--- a/keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
+++ b/keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
@@ -43,6 +43,12 @@ class SecretValueConverterUtilTest {
   }
 
   @Test
+  void emptyLineTerminationsReturnsEmptyStrings() {
+    assertThat(mapSecretValue((k, v) -> v, "key", "\n")).containsOnly("");
+    assertThat(mapSecretValue((k, v) -> v, "key", "\nok\n\n")).containsOnly("", "ok");
+  }
+
+  @Test
   void nullMappedIsNotReturned() {
     Set<String> mappedValues =
         mapSecretValue(
--- a/keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
+++ b/keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2023 ConsenSys AG.
+ *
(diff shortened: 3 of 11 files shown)


keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class SecretValueConverterUtil {
```
siladu, line 24: nit: naming - probably don't need Util. 
'Convertor' in class, 'map' in method - IMO either SecretValueMapper or convertSecretValue

siladu: > probably don't need Util.

On second thoughts, I guess this is a Util for the mapper function itself, so maybe makes sense

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
```diff
@@ -0,0 +1,52 @@
...
+  private static final Logger LOG = LogManager.getLogger();
+
+  public static <R> Set<R> mapSecretValue(
+      BiFunction<String, String, R> mapper, String propKey, String secretValue) {
+    final AtomicLong discardedValues = new AtomicLong(0);
```
siladu, line 29: Why does this local var need to be Atomic?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
```diff
@@ -0,0 +1,76 @@
...
+class SecretValueConverterUtilTest {
+
+  @Test
+  void singleValueIsMapped() {
+    Set<String> mappedValues = mapSecretValue((k, v) -> v, "key", "value");
+    assertThat(mappedValues).containsOnly("value");
+  }
+
+  @Test
+  void newlinesValuesAreMapped() {
```
gfukushima, line 31: Maybe a test case where a possibly "buggy" secret value is store with a starting \n? If that makes sense

gfukushima (approved):
lgtm

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
siladu: Is it more of a SecretValueExtractor? Maybe could separate the concerns of extracting and then do the mapping afterwards, or would that be poor performance?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
usmansaleem: I'll add one ... it'll return an empty string, which will fails in our actual mapper in Web3Signer which uses BLS key mapping to return null.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
```diff
@@ -0,0 +1,52 @@
...
+                    discardedValues.incrementAndGet();
+                  }
+                  return obj;
+                })
+            .filter(Objects::nonNull)
+            .collect(Collectors.toSet());
+
+    if (discardedValues.get() > 0) {
+      LOG.warn(
+          "AWS Secrets - {} values from property {} were not mapped and discarded.",
```
siladu, line 46: Is there value in knowing which values were discarded?

keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
siladu: Along the same line of thinking, what if the file has lots of new lines?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
usmansaleem: no .... since they can technically be private keys (or some combination of) ... don't really want to log the values.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
usmansaleem: @siladu just pushed a test case to cover it ... basically all empty `\n` will be converted to empty strings .. which in case of web3signer will return null because BLS private key conversion will fail.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
usmansaleem: out of habit I guess ... since the stream can be multithreaded. In this case I agree that we can use standard long value.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
jframe, line 24: Can move to a common package as this isn't specific to AWS. We can reuse the secret conversion for Hashicorp and Azure

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
jframe: Definitely should not log values.

Knowing which index out of a large list of values would be helpful though to help pinpoint the bad key though.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
```diff
@@ -0,0 +1,82 @@
...
+package tech.pegasys.signers.aws;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static tech.pegasys.signers.aws.SecretValueConverterUtil.mapSecretValue;
+
+import java.util.Set;
+
+import org.junit.jupiter.api.Test;
+
+class SecretValueConverterUtilTest {
```
jframe, line 22: Test case for separating lines by carriage return as well? Or do we only want to support \n?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
siladu: Key + index then?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
usmansaleem: moved to common module under keystore.

keystorage/aws/src/test/java/tech/pegasys/signers/aws/SecretValueConverterUtilTest.java
usmansaleem: since we are using `String.lines()`, it technically support all kind of line termination.

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
```diff
@@ -26,27 +26,22 @@ public class SecretValueConverterUtil {
...
     final Set<R> mappedValues =
         secretValue
             .lines()
             .map(
                 v -> {
+                  long index = valuesIndex.getAndIncrement();
                   final R obj = mapper.apply(propKey, v);
                   if (obj == null) {
-                    discardedValues.incrementAndGet();
+                    LOG.warn("Secret Value {}:{} was not mapped and discarded.", propKey, index);
```
siladu: I would read "Secret Value" as the secret value.

Suggest "Secret key {} with index {} ..."?

siladu: Does secret key mean anything to the users, from a user's point of view, what actually is the secret key with this multi-secret setup?

siladu: "multi-part key"?

gradle/versions.gradle
```diff
@@ -137,5 +137,8 @@ dependencyManagement {
        \-- org.apache.tuweni:tuweni-net:2.3.1
      */
     dependency 'commons-net:commons-net:3.9.0'
+
+    // used in tests to assert log message
+    dependency 'de.neuland-bfi:assertj-logging-log4j:0.5.0'
```
siladu, line 142: Nice solution!
Should we limit this to test scope in the deps?

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
siladu: @usmansaleem I think you missed this change

keystorage/aws/src/main/java/tech/pegasys/signers/aws/SecretValueConverterUtil.java
usmansaleem: Now that this code is in common module, the verbage has to be common enough to work for AWS and Azure etc. `Secret key {} with index {}` ... make sense to me.

gradle/versions.gradle
usmansaleem: I don't think we can limit the dependency declaration to test or other scopes here. It is declared as testImplementation in common's build.gradle though.

siladu (approved):
LGTM

keystorage/common/src/main/java/tech/pegasys/signers/common/SecretValueMapperUtil.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.common.collect.Streams;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class SecretValueMapperUtil {
+  private static final Logger LOG = LogManager.getLogger();
+
+  public static <R> Set<R> mapSecretValue(
+      BiFunction<String, String, R> mapper, String secretName, String secretValue) {
+    return Streams.mapWithIndex(
```
siladu, line 29: nice 👍

siladu (approved):
LGTM

PR #177, opened 2023-02-27
Suppress false positive CVE-2022-45688 for json-java

usmansaleem:
Update owasp gradle plugin version
Update AWS to latest minor version
Add suppression for json-java


siladu (approved):
LGTM

PR #183, opened 2023-03-23
Fix AWS Acceptance test assertion

usmansaleem:
Fix AWS Acceptance test assertion. Since CI runs the test against live AWS secrets manager, some of the keys return fails due to authentication error. Hence fix the errorCount to not assert on fix number.


siladu (approved):
LGTM

PR #184, opened 2023-03-28
Azure map secrets tag filter

usmansaleem:
Azure bulk map secrets can now be filtered based on tags.

### DIFF
--- a/keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
+++ b/keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
@@ -14,13 +14,12 @@ package tech.pegasys.signers.azure;
 
 import tech.pegasys.signers.common.MappedResults;
 
-import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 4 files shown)


siladu (approved):
LGTM

keystorage/azure/src/main/java/tech/pegasys/signers/azure/AzureKeyVault.java
```diff
@@ -149,4 +152,14 @@ public <R> MappedResults<R> mapSecrets(final BiFunction<String, String, R> mappe
...
   }
+
+  private static boolean secretPropertiesPredicate(
+      final Map<String, String> tags, final SecretProperties secretProperties) {
+    if (tags == null || tags.isEmpty()) {
+      return true; // we don't want to filter if user-supplied tags map is empty
+    }
+
+    return secretProperties.getTags() != null // return false if remote secret doesn't have any tags
+        && secretProperties.getTags().entrySet().containsAll(tags.entrySet());
```
jframe, line 163: This requires matching both the key and value, which is slightly different from what was done for AWS. For AWS it could either be keys, values or both keys and values that would be matched. Don't remember why we did that though.

jframe, line 163: So the secret client doesn't filters?

jframe (approved):

